PR #7, opened 2019-09-26
Add OAuth login flow with auth guard and login UI to CRBM-Viz

gmarupilla:
TODO: Show info of currently logged in user in the corner of nav bar


CRBM-Viz/angular.json
```diff
@@ -28,6 +28,7 @@
               "src/assets"
             ],
             "styles": [
+              "./node_modules/bootstrap/dist/css/bootstrap.min.css",
```
bilalshaikh42: I would like to stick to using angualr material only, without pulling in other styles? Would that work?

CRBM-Viz/src/app/app.module.ts
```diff
@@ -52,10 +60,31 @@ const declarations = [
 ];
 
 //Services
-import { VisualizationsService } from './Services/visualizations.service';
-import { SimulationService } from './Services/simulation.service';
-import { AuthService } from './Services/auth.service';
-const providers = [VisualizationsService, SimulationService, AuthService];
```
bilalshaikh42: Could you edit this portion to ensure that the visualization and simulation services are imported correctly as well?

bilalshaikh42 (changes requested):

bilalshaikh42 (approved):

Issue #9, opened 2019-09-26
Tests are failing 


bilalshaikh42:
Need to write tests for each component

bilalshaikh42:
@gmarupilla Could you work on this as well? We need to get every test passing atleast for the unit tests. As we add more components it will get harder to manage so we should get this done before any further work

gmarupilla:
Yeah! Sure.

gmarupilla:
I looked into it. The problem with writing tests right now is that the requirement is not very concrete as of now. If i write tests right now, they'll have to be changed again once we start developing according to OAS file. What do you think?

bilalshaikh42:
We should definitely have the tests set up even though the spec is not finalized. 

Firstly for unit tests, the majority of the test cases will not be dependent on the structure of objects but tests more basic features like ensuring that a list is properly populated or that the links and routing work. The ones that do rely on the objects _Should_  fail if the spec changes so that we know that the front end needs updating.
 
Secondly, the e2e integration tests should not be dependent on the spec at all, but instead ensure that the site is working, regardless of what the api is sending out. Is a spec change breaks the e2e tests then that is a sign that the site needs to be adapted to the new spec, not that the tests must be changed.

bilalshaikh42:
@gmarupilla 
All the tests have been fixed except for 2, the vega-viewer component and the file-edit component. 
Both are giving an error about a parameter being null. For both components, the parameter is set using an HTTP Request. 
We need to follow the below directions to mock the HTTPS request and give back a test parameter for the components to render. 

https://blog.angulartraining.com/how-to-write-unit-tests-for-angular-code-that-uses-the-httpclient-429fa782eb15

bilalshaikh42:
fixed partially by #35 
New issues opened for individual tests failing

PR #26, opened 2019-10-09
Wire Auth0 token interceptor and file edit/delete flow in CRBM-Viz

gmarupilla:
I added client id and domain to config and kept it in gitignore. I've added config.example.json which the user can copy paste as config.json and modify the contents for testing purposes. Or environment wise.


bilalshaikh42:
The build and docker tests are failing. Could you make sure that the app builds properly outside your local environment? 
@gmarupilla

gmarupilla:
I've added `config.example.json` as a default config.. Or a template config..every time we want to build it.. We will copy example config to `config.json`
And then it'll work. It's a replacement for environment variable. Since in front-end app we cannot access environment variables

bilalshaikh42:
> I've added `config.example.json` as a default config.. Or a template config..every time we want to build it.. We will copy example config to `config.json`
> And then it'll work. It's a replacement for environment variable. Since in front-end app we cannot access environment variables

I see that the config file contains just the Auth0 client id. In that case, it does not need to be hidden and should be included in the source to allow for automatic building and deployment

https://community.auth0.com/t/client-id-vs-secret/9558

gmarupilla:
> > I've added `config.example.json` as a default config.. Or a template config..every time we want to build it.. We will copy example config to `config.json`
> > And then it'll work. It's a replacement for environment variable. Since in front-end app we cannot access environment variables
> 
> I see that the config file contains just the Auth0 client id. In that case, it does not need to be hidden and should be included in the source to allow for automatic building and deployment
> 
> https://community.auth0.com/t/client-id-vs-secret/9558

Still, I think it's better to have such things in a configuration file rather than being hardcoded into code itself. Here's one use case why, I have my own personal auth0 account which I use to experiment/test the code that I write, if I modify client id in code, I'll need to exempt it from commits every single time I change anything else. With a configuration file like this, user can modify configuration with ease according to different environments. To resolve this, I added docker rule to copy config.example.json (which is default config for heroku) to config.json. Moreover, more things are also added to config, and will keep on increasing in the future.

bilalshaikh42:
> > > I've added `config.example.json` as a default config.. Or a template config..every time we want to build it.. We will copy example config to `config.json`
> > > And then it'll work. It's a replacement for environment variable. Since in front-end app we cannot access environment variables
> > 
> > 
> > I see that the config file contains just the Auth0 client id. In that case, it does not need to be hidden and should be included in the source to allow for automatic building and deployment
> > https://community.auth0.com/t/client-id-vs-secret/9558
> 
> Still, I think it's better to have such things in a configuration file rather than being hardcoded into code itself. Here's one use case why, I have my own personal auth0 account which I use to experiment/test the code that I write, if I modify client id in code, I'll need to exempt it from commits every single time I change anything else. With a configuration file like this, user can modify configuration with ease according to different environments. To resolve this, I added docker rule to copy config.example.json (which is default config for heroku) to config.json. Moreover, more things are also added to config, and will keep on increasing in the future.

Got it. In that case, lets make use of the environment.ts and environment.prod.ts files in the environments folder. The auth service has a good example on how I used the files to change based on various environments. You could even add a new environment for your development environment. Thi way, everything needed to build the app (currently the build step is failing in the node workflow) is present in default configurations at all times. When building, use the --configuration flag
Here is a good guide for it:
[Link]( https://blog.angularindepth.com/becoming-an-angular-environmentalist-45a48f7c20d8)

gmarupilla:
> > > > I've added `config.example.json` as a default config.. Or a template config..every time we want to build it.. We will copy example config to `config.json`
> > > > And then it'll work. It's a replacement for environment variable. Since in front-end app we cannot access environment variables
> > > 
> > > 
> > > I see that the config file contains just the Auth0 client id. In that case, it does not need to be hidden and should be included in the source to allow for automatic building and deployment
> > > https://community.auth0.com/t/client-id-vs-secret/9558
> > 
> > 
> > Still, I think it's better to have such things in a configuration file rather than being hardcoded into code itself. Here's one use case why, I have my own personal auth0 account which I use to experiment/test the code that I write, if I modify client id in code, I'll need to exempt it from commits every single time I change anything else. With a configuration file like this, user can modify configuration with ease according to different environments. To resolve this, I added docker rule to copy config.example.json (which is default config for heroku) to config.json. Moreover, more things are also added to config, and will keep on increasing in the future.
> 
> Got it. In that case, lets make use of the environment.ts and environment.prod.ts files in the environments folder. The auth service has a good example on how I used the files to change based on various environments. You could even add a new environment for your development environment. Thi way, everything needed to build the app (currently the build step is failing in the node workflow) is present in default configurations at all times. When building, use the --configuration flag
> Here is a good guide for it:
> [Link](https://blog.angularindepth.com/becoming-an-angular-environmentalist-45a48f7c20d8)

Thanks for the info, moved configs to environment files.

gmarupilla:
I guess tests are failing in the components you've written. Can you look into it?

bilalshaikh42:
Yup tests are not yet passing per #9 .  For now we can only assume it works if building. I'll merge this in

CRBM-Viz/angular.json
```diff
@@ -63,6 +63,35 @@
...
               ]
+            },
+            "development": {
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.dev.ts"
+                }
+              ],
+              "optimization": true,
```
bilalshaikh42: This might be very slow for development. Up to you, but we can change the optimization and AOT off since we dont be deploying using this environment

CRBM-Viz/src/app/Components/loading-spinner/loading-spinner.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="container">
+    <div class="row">
+        <div class="span12" style="height: 80vh; padding-top: 30vh; padding-left: 30vw;">
```
bilalshaikh42: Can you move the styles to the sass? If there is a reason for it to be here, its fine in that case

CRBM-Viz/src/app/Components/loading-spinner/loading-spinner.component.sass
```diff
@@ -0,0 +1,49 @@
+@keyframes lds-eclipse
```
bilalshaikh42: Angular material already has a spinner defined. 
Can we use that instead for simplicity and consistancy? 
https://material.angular.io/components/progress-spinner/overviewhttps://material.angular.io/components/progress-spinner/overview

CRBM-Viz/src/app/Layout/links/links.component.ts
```diff
@@ -8,5 +9,13 @@ import { AuthService } from 'src/app/Services/auth0.service';
 export class LinksComponent implements OnInit {
   constructor(public auth: AuthService) {}
 
-  ngOnInit() {}
```
bilalshaikh42: As mentioned in #29 this component will be deleted, so it is not a good idea to have the profile load here. In any case, it is probably not a good idea to have the profile load in an unrelated component. Can this be moved to the main app component?

CRBM-Viz/src/app/Services/auth0.service.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { tap, catchError, concatMap, shareReplay } from 'rxjs/operators';
 import { Router } from '@angular/router';
 import { environment } from 'src/environments/environment';
+import { environment as config }  from 'src/environments/environment';
```
bilalshaikh42: This is just messy. Lets change the config name below to environment

CRBM-Viz/src/app/Services/auth0.service.ts
```diff
@@ -22,8 +23,8 @@ export class AuthService {
   redirect = this.returnTo + 'callback';
   auth0Client$ = (from(
     createAuth0Client({
-      domain: 'crbm.auth0.com',
-      client_id: '0NKMjbZuexkCgfWY3BG9C3808YsdLUrb',
+      domain: config.auth0.domain,
```
bilalshaikh42: Change this to environment.autho0

CRBM-Viz/src/app/Services/auth0.service.ts
```diff
@@ -121,9 +140,10 @@ export class AuthService {
     this.auth0Client$.subscribe((client: Auth0Client) => {
       // Call method to log out
       client.logout({
-        client_id: '0NKMjbZuexkCgfWY3BG9C3808YsdLUrb',
+        client_id: config.auth0.clientId,
```
bilalshaikh42: Likewise here

bilalshaikh42 (changes requested):

PR #66, opened 2019-10-21
Data

bilalshaikh42:
Add components and page for displaying tabular data


bilalshaikh42:
@gmarupilla You can go ahead and merge this in if you agree with changes!
It provides a new data-table component that can be used for any situation where we need to display tabular data. It might be worth switching this over to the built-in data table, but that will take some work as we would need to build all the included features manually.

gmarupilla:
> @gmarupilla You can go ahead and merge this in if you agree with changes!
> It provides a new data-table component that can be used for any situation where we need to display tabular data. It might be worth switching this over to the built-in data table, but that will take some work as we would need to build all the included features manually.

Yes! I could definitely merge and you can switch to mat-tables later.

PR #71, opened 2019-10-22
Add sidenav navigation, custom spinner, and pagination fixes to CRBM-Viz

gmarupilla:


CRBM-Viz/src/environments/environment.ts
```diff
@@ -2,9 +2,9 @@ export const environment = {
   production: false,
   baseUrl: '/',
   auth0: {
-    domain: 'crbm.auth0.com',
-    clientId: '0NKMjbZuexkCgfWY3BG9C3808YsdLUrb',
-  },
```
bilalshaikh42: Please dont edit the default the environment! Can you use dev for this, or is there a particular reason you need it to be the default?

bilalshaikh42 (changes requested):

PR #104, opened 2019-10-30
adding local deployment instructions and other info to README

bilalshaikh42:


CRBM-Viz/angular.json
```diff
@@ -7,7 +7,8 @@
       "projectType": "application",
       "schematics": {
         "@schematics/angular:component": {
-          "style": "sass"
+          "style": "sass",
+          "styleext": "scss"
```
bilalshaikh42: This will break the styles for all the components that currently use sass. Perhaps the theme can be changed to sass instead of scss?

CRBM-Viz/src/app/Services/auth0.service.ts
```diff
@@ -44,6 +44,23 @@ export class AuthService {
   // Create subject and public observable of user profile data
   private userProfileSubject$ = new BehaviorSubject<any>(null);
   userProfile$ = this.userProfileSubject$.asObservable();
+  extendedProfile: object = new Object({
```
bilalshaikh42: The type of this object must be specified in a model and used here. The object type wont have any of the feilds added here, leading to the compile error currently

CRBM-Viz/angular.json
bilalshaikh42: Nevermind. When merged, this seems to work

CRBM-Viz/src/app/Layout/navigation/navigation.component.html
```diff
@@ -1,4 +1,4 @@
-<mat-sidenav-container class="sidenav-container">
+<mat-sidenav-container class="sidenav-container" hasBackdrop="true">
```
bilalshaikh42: I think this should stay as is. the backdrop is conditional on the screensize. This adds a backdrop even on wide screens, where the menu and content can fit side by side

CRBM-Viz/angular.json
jonrkarr: The components can still use sass. I changed to use a custom theme rather than the default theme. The custom theme is in scss.

CRBM-Viz/src/app/Layout/navigation/navigation.component.html
jonrkarr: Fine with me.

CRBM-Viz/src/app/Services/auth0.service.ts
jonrkarr: That makes sense. Having not ready any of the documentation for Angular, I'm slowly learning.

You could leave this out until I create a model.

jonrkarr:
I'm replacing this with a new pull request for the latest revision to this branch

PR #106, opened 2019-10-31
Modules

bilalshaikh42:
This is a reorganization of the app into separate modules to allow for lazy loading of the various sections (Visualize, Simulate, Profile, Content, etc.) 

I am going to hold off on the profile and simulate modules until pull requests #104 and #71 are completed, as it may cause some conflicts due to the files moving.  Until then, those components will be included into the share module


gmarupilla:
Why lazy loading? Angular is a Single Page application, so ideally, everything related to app will be (and should be) loaded in one go. If not, we can show a loader until everything is loaded into browser. After that, every view will be dependent on request and response from the back-end. For which, again we're planning to write a loader, implement pagination for tabular data. So where do we want lazy loading? Even if we do, I don't think any file structure reorganization will help. Angular just doesn't work that way.

Also, please revert these changes, because of this change I'm having merge conflicts in my PR. It's just simple reorganization, can be merged later as well after #71 is merged. But resolving conflicts for me is an additional and unnecessary overhead. And I'd really appreciate if you do not modify any code related to #71 because, first, you're modifying the old code, which will be replaced by the code in already existing PR. Second, I don't want to resolve the conflicts every time I want to push a minor change.

PR #132, opened 2019-11-06
Layout style

jonrkarr:
- Fix top bar for Firefox
- Spacing out home page image


jonrkarr:
Yes, its set to merge. 

The simulation grid is working and the grid is factored out to its own component. I'm going to re-use it to display the status of simulations owned by the user and for the model module.

There's just one issue I haven't figured out. If you can figure it out, that would be great. 
- The styles for the grid are in `src/styles.scss` (organized into the `.grid` class). They should be in the grid component (`src/app/Shared/Components/grid.sass`). For reasons I don't understand, the styles weren't active when I tried to move them to the component.

bilalshaikh42:
>     * grid

I'll take a look at that. 

One issue that I noticed: 
When multiple columns are selected in the table, they are added without enabling horizontal scrolling within the table. If too many are added, then each column resizes to be too small, and cant be read. if there was a minimum size for each column equal to the text length and horizontal scrolling was used a user could scroll to see all the columns

jonrkarr:
The minimum width of the columns is now set based on the width of the table and the number of active (non-hidden) columns.

bilalshaikh42:
I am still seeing the issue. For example, here is the chart with all the columns added. much of the text is not visible
image

jonrkarr:
I think the min width is working now.

bilalshaikh42:
Still seeing it. Only slightly off with the sidebar closed but for so with it open.

image

image

bilalshaikh42:
We can adress that later though. I'll merge this in

PR #163, opened 2019-11-19
Layout style

jonrkarr:
- Styling
- Working on models, services, and pipes

Already merged with master


bilalshaikh42:
The profile page is only showing a blank profile picture. 
Is this intended behavior?

bilalshaikh42:
> The profile page is only showing a blank profile picture.
> Is this intended behavior?

I see, there is a user id for the profile. 
We need to have the current signed in user be the defalt profile. 
Ill create an issue for that.

PR #165, opened 2019-11-19
Layout style

jonrkarr:
- Adding private access token to visualization model
- Adding fields to track authors separate from owners. 
  - `authors` is a list of `Person` or `User`. 
  - This will enable us to track users in author lists.
- Simplifying static methods in UtilsService and removing injection of UtilsService
- Removing deprecated ng-deep styling; moving to `src/styles.scss`
- Fixing scrolling of left menu
- Renaming `profile` route to `user`, renaming `account` module to `user`
- Adding user name to crumb bar


jonrkarr:
I used a third-party component because it didn't look like that material component accepted data. After looking through the documentation details, I see it does accept data. I can switch to the material component.

jonrkarr:
I re-implemented the history tree with the material tree component.

PR #201, opened 2019-12-05
add user management

bilalshaikh42:


bilalshaikh42:
The login component now calls the API to ensure that a copy of the user ID exists. 
The profile component makes a request to the API for the user profile. 
Auth0 registration now requires choosing a username. If social sign on is used username is  received from social profile

bilalshaikh42:
@jonrkarr  could you confirm that the current changes do not break the profile demo?

bilalshaikh42:
> 
> 
> @jonrkarr could you confirm that the current changes do not break the profile demo?

This is confirmed to work. When the api is running, a one line change will pull non-demo data

bilalshaikh42:
Build is failing since other services depend on removed getuserbyauthid() method.
Potential solutions:

1) Other services get username from authentication service, then call user service on that username
-  A little bit more work per service, but maintains good separation of concerns
- Username might not be most secure method of checking access 

2)  User service returns current user if no username is provided
- User service will have dependency on auth service
- User profile can be cached as observable (check how the auth service does this, I think it provides caching by sending the same observable object every time) 

3) Auth service includes method for current user profile 
- objectively bad idea.  Auth service will then depend on user service, user model and create circular dependency 

@jonrkarr  It seems that the method is used for providing a "default" route parameter wherever a username is expected. This is essentially the same issue that is solved in the refactored profile component. 
Do you think it would make sense to move that logic into the user service to have it be reusable? The issue with this setup is that the root level urls will now depend on the user being logged in. This makes sense for the "/user" route, but not for the eg. "/models" route. 
I believe the current set up is  essentially routing "/models" to "your models" when it should be routing it to browse models

jonrkarr:
I'll look into this.

`getByAuth0Id` is used in the user browse projects/models/simulations/visualizations components. We don't need the user profile. We just need to get the list of projects/models/simulations/visualizations owned by the user. There's a todo for this. I've ignored implementing this.

jonrkarr:
The tests should be passing now. I implemented out the management of resource ownership slightly further. The retrieval of projects/models/simulations/visualizations owned by a user is now moved to the services. My intension was to put that functionality in the services (more specifically the API behind the services). I largely haven't gotten that far.

jonrkarr:
All set to merge with master

PR #463, opened 2020-04-09
Datamodel interfaces and DTO's

bilalshaikh42:
Changes to data model to help with abstraction and code reuse. 
Alignment with SBML


jonrkarr:
I made several edits. They're pushed to `datamodel-karr`.

jonrkarr:
I just added 1 more small addition to the Algorithm data model.

PR #1066, opened 2020-08-24
Dev dispatch

jonrkarr:
Styling submission form in dispatch app. 

Still todo to style dispatch app
- Style simulation results page


jonrkarr:
@bilalshaikh42 can you fix the lint issue in `apps/dispatch/src/app/components/dispatch/dispatch.component.htmlapps/dispatch/src/app/components/dispatch/dispatch.component.html`?

jonrkarr:
The dispatch UI is now cleaned up except for styling Plotly better.

Please review and merge. I tried to test the visualization as best as possible without being able to use the API, using mocked data.

bilalshaikh42:
> @bilalshaikh42 can you fix the lint issue in `apps/dispatch/src/app/components/dispatch/dispatch.component.htmlapps/dispatch/src/app/components/dispatch/dispatch.component.html`?

The linting rules say that this component should be refactored. I can disable this for now and create an issue to fix this, but we should be mindful of letting these build up

PR #1078, opened 2020-08-26
Add async results API for visualizations and refactor dispatch IO

### LINKED ISSUE #782: Create an API to run simulations
Due to requirements of NAR Webserver issue, we need an API that is not restricted to users with accounts. 
The "dispatch" portion of the system can meet this requirement. The API should meet the following base requirements (work in progress): 

1. Have routes to accept a simulation through an OMEX archive.
   a. Generate a unique id for this run
   b. Save the needed files somewhere (db/fs/hpc)
   c. Send a message to the dispatch service about running the simulations 
   d. Return to the user the id of the dispatched simulation

2. Have a route to get the results of a simulation run based on the  id provided
 
Later enhancements
----------------------------------------------------------------------------------------

3.  Have a route to run a simulation based on the Biosimulations ID

4  Allow users to optionally login- get a higher priority dispatch/save to a more persistent id

gmarupilla:
Fixes #782


biosimulations/apps/dispatch-api/src/app/app.controller.ts
```diff
@@ -138,102 +127,66 @@ export class AppController implements OnApplicationBootstrap {
...
-              .pipe(fs.createWriteStream(jsonPath));
-          }, 0);
-
-          const fileContent = fs.readFileSync(jsonPath);
-          jsonResults[directoryName][taskName] = JSON.parse(
-            fileContent.toString()
-          );
+    sedmls.forEach((sedml) => {
+      structure[sedml] = [];
+      const taskFiles = fs.readdirSync(path.join(resultPath, sedml));
```
bilalshaikh42: Could you change these to use an async method? The sync methods will only allow the API to  handle one client at a time

biosimulations/apps/dispatch-api/src/app/app.controller.ts
```diff
@@ -138,102 +127,66 @@ export class AppController implements OnApplicationBootstrap {
...
+    const jsonPath = path.join(
+      fileStorage,
+      'simulations',
+      uId,
+      'out',
+      sedml,
+      task
+    );
+    const filePath = chart ? `${jsonPath}_chart.json` : `${jsonPath}.json`;
+    const fileContent = JSON.parse(fs.readFileSync(filePath).toString());
```
bilalshaikh42: Same here, we need to switch all of these to async

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-api/src/app/app.controller.ts
gmarupilla: I can, but by default the async method for file read expects a callback which is fine for a non API use case, we need to return the data as JSON in response which by itself is a synchronous use case in my opinion, do you want me to wrap the async call in a promise and then return that promise? This is the only way that I think which we can make it async.

But internally it would actually by synchronous because NestJS will wait for the promise to get resolved before returning the response, or that won't be the case and I'm missing something here?

bilalshaikh42: So while nest will resolve the promise, it will be able to handle other user requests. 
Here is a look at how to do this using promised. Essentially just using the await keyword will have node do all the work for you . 
https://nodejs.org/api/fs.html#fs_promise_example

gmarupilla: Done

biosimulations/apps/dispatch-api/src/app/app.controller.ts
gmarupilla: Done

bilalshaikh42 (changes requested):
Hi akhil,
You are still using synchrnous methods at various locations (such as line 105 in the upload method). Could you change those?

biosimulations-daemon (dismissed):
Each iteration of this loop will be blocking due to  the await. A observable/promise followed by a map would be the better approach here

bilalshaikh42:
Lets hold off on creating a library for the utils on this PR. If we create a library, then then nx system starts to diverge and becomes very difficult to integrate. Abstracting the library out can be its own future pr, some code repetition is fine for now

gmarupilla:
I removed the job monitoring code, added a new API (/api/dispatch-finish) to trigger chart data generation. The chart data won't be generated unless and until the mentioned API is hit.

bilalshaikh42 (approved):

PR #1144, opened 2020-09-04
Simulators api

bilalshaikh42:
Adds api for viewing biosimulators and their metadata


bilalshaikh42:
Still pending: Using the api for ontology terms, still using the json files
The database needs to be updated with more of the simulators, currently only contains vcell and copasi

PR #1156, opened 2020-09-08
Dev config

jonrkarr:
Re-applying changes that didn't get merged into dev.
- Management for app-specific configuration -- implemented similarly to auth config
- More documentation
- Edited documentation


bilalshaikh42 (changes requested):
There is a library in the config folder called config-angular that is intended to do what shared-services seems to be doing. 
It will be using the config-shared library to keep the urls in sync with the back end, so that they do not need to be defined twice. It is also intended to use the env files to determine if the app is running locally so we can test against a local api. 

Can we continue to use the workaround  of defining the settings for now?

jonrkarr:
Environment variables have a similar, but distinct purpose. Some recommend separate these two types of configuration. While this configuration has URLs, these URLs are not for the purpose of specifying the locations of services (e.g., where the application should obtain data). These URLs are for the purposes of providing content for documentation.

The angular config module is empty. If you want to replace the angular config with my service config, that would be fine.

jonrkarr:
Let's merge this, and you can refactor later. At least this introduces a structured representation of the config values, clearly outlines the places where the values are needed, and sets up accessing the values from a centralized source.

bilalshaikh42:
I'll work on moving the service. Another issue is with the dependency tree.  The shared-ui library cannot have a dependency on a service as it then needs to be aware of the service structure. The ui components should only be using inputs to keep them purely functional. The service injection needs to happen further up the dependency tree.

jonrkarr:
If you want to change the dependency tree, this will require passing configuration down through components. E.g.,
- app.component.ts
  - navigation.component.ts
    - logo-image.component.ts
    - logo-text.component.ts
    - privacy-policy-notice.component.ts

The shared content is similar.
- app/help/help/help.component.ts, app/help/faq/faq.component.ts
  - text-page-feedback-side-bar-section.component.ts
  - text-page-help-side-bar-section.component.ts

jonrkarr:
Merged with recent changes to integrated into `dev`

jonrkarr:
I refactored shared-ui to remove the dependency on ConfigService.

This branch needs to be merged in so that the applications have the correct logos and logo text. Refactoring will need to wait until a later time.

PR #1172, opened 2020-09-10
Add cron-based HPC job monitoring and Mongo persistence to dispatch service

### LINKED ISSUE #1124: Status monitoring for Jobs
A proposed approach for monitoring jobs: 

The jobs are run on the HPC as following: 

1. The sbatch template creates a file called STARTED in the directory, which contains some basic info about the job
2. The job runs as normal, outputting to the directory
3. If there is an error in the simulation it is output into a file called ERROR
4. When the job completes, the sbatch template outputs a file called DONE with some basic timing info

The jobs are monitored as following: 
1. The dispatch service maintains an intenal state of the simulation that starts in "QUEUED" stated.
2. It then  uses the [ file system watch method](https://nodejs.org/docs/latest/api/fs.html#fs_fs_watch_filename_options_listener) to watch the directory for the presence of the STARTED, ERROR, and DONE files
3. Upon any change to those files, it uses the callback to update its internal state.
4.  When the state is updated, it sends a message to the message system that alerts the system that the results can then be processed by other apps (for example, this is how the results can be generated)

gmarupilla:
Fixes #1124


biosimulations/apps/dispatch-api/src/app/app.module.ts
```diff
@@ -7,14 +7,21 @@ import {
   NatsOptions,
 } from '@nestjs/microservices';
 import { AppController } from './app.controller';
-import { AppService } from './app.service';
 import { ConfigService } from '@nestjs/config';
 import { BiosimulationsConfigModule } from '@biosimulations/config/nest';
+import { MongooseModule } from '@nestjs/mongoose';
+import { ScheduleModule } from '@nestjs/schedule';
+
+const dbUri = process.env.MONGO_URI || '';
```
bilalshaikh42: This won't work on the docker image. You need to use the config service to import the uri

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -19,16 +19,26 @@ import { SimulationDispatchSpec } from '@biosimulations/dispatch/datamodel';
...
 export class AppController {
   constructor(
     private readonly configService: ConfigService,
     private hpcService: HpcService,
     private sbatchService: SbatchService,
-    @Inject('DISPATCH_MQ') private messageClient: ClientProxy
+    @Inject('DISPATCH_MQ') private messageClient: ClientProxy,
+    private schedulerRegistry: SchedulerRegistry
   ) {}
   private logger = new Logger(AppController.name);
```
bilalshaikh42, line 41: Both of these values will be the same instance if the logger. Is that the intention?

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -130,15 +144,39 @@ export class AppController {
...
+    this.jobMonitorCronJob(slurmjobId, uuid, 30);
+  }
+
+  async jobMonitorCronJob(jobId: string, uuid: string, seconds: number) {
+    const job = new CronJob(`${seconds.toString()} * * * * *`, async () => {
+      const squeueRes = await this.hpcService.squeueStatus(jobId);
+      const jobMatch = squeueRes.stdout.match(/\d+/);
+      const isJobRunning = jobMatch !== null && jobMatch[0] === jobId;
+      if (!isJobRunning) {
+        this.messageClient.emit('dispatch_finish', { jobId, uuid });
```
bilalshaikh42: Not sure if this in elsewhere, but we need to end the job at this point correct?

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-service/src/app/app.controller.ts
bilalshaikh42: I am a bit confused here. Are you intending to use the message system to trigger a call within the same service?

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: Yes, instead of passing the control of job callback to another method, I thought it'd be better to just notify NATS and end the flow of control of job there only. Afterwards, whichever controller wants to get notified when any simulation finishes, it can just subscribe to this message. 
For now, we only have 1 subscription for this message to start processing the results into JSON.

bilalshaikh42:
The cron job would need to be stopped at this point however. Also, the
method receiving the message cannot be stopping the cron job. There is no
guarantee that the deployment receiving the message will be the same as the
deployment that sent it. If there are replicas of the service then the
another instance will recover the message. That instance would not have the
job running and crash when attempting to stop it.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -130,15 +140,40 @@ export class AppController {
...
+  }
+
+  async jobMonitorCronJob(jobId: string, uuid: string, seconds: number) {
+    const job = new CronJob(`${seconds.toString()} * * * * *`, async () => {
+      const squeueRes = await this.hpcService.squeueStatus(jobId);
+      const jobMatch = squeueRes.stdout.match(/\d+/);
+      const isJobRunning = jobMatch !== null && jobMatch[0] === jobId;
+      if (!isJobRunning) {
+        this.messageClient.emit('dispatch_finish', uuid);
+        this.schedulerRegistry.getCronJob(jobId).stop();
```
gmarupilla, line 168: @bilalshaikh42 That part. I already fixed that 3 commits back, maybe you're seeing the code from previous commits? Can you check if the issue that you're talking about is still there in the latest commits?

bilalshaikh42: Yes, this was in reference to the commit [here](https://github.com/biosimulations/Biosimulations/pull/1172#discussion_r488135890)

biosimulations/jest.config.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   resolver: '@nrwl/jest/plugins/resolver',
   moduleFileExtensions: ['ts', 'js', 'html'],
   coverageReporters: ['lcov'],
+  testEnvironment: 'node',
```
bilalshaikh42: This will cause issues for the Angular apps. 
You can set this options for the apps individually

bilalshaikh42 (changes requested):

biosimulations/jest.config.js
gmarupilla: Yes, I'll remove from here.

bilalshaikh42 (dismissed):

biosimulations/apps/dispatch-service/src/app/app.controller.spec.ts
```diff
@@ -137,5 +140,15 @@ describe('AppController', () => {
     });
   });
 
-  
+  describe('test jobMonitorCronJob', () => {
+    it('should return promise', () => {
+      const appController = app.get<AppController>(AppController);
```
bilalshaikh42, line 146: This test is going to break if the job with that id is removed from the cluster.

biosimulations/nx.json
```diff
@@ -133,6 +133,9 @@
     },
     "dispatch-datamodel": {
       "tags": ["scope:dispatch", "type:datamodel", "platform:any"]
+    },
+    "dispatch-api-models": {
+      "tags": ["scope:dispatch", "type:databasemodel", "platform:any"]
```
bilalshaikh42: need to change type

biosimulations/angular.json
```diff
@@ -1406,6 +1406,33 @@
...
+    },
+    "dispatch-api-models": {
+      "root": "libs/dispatch/api-models",
+      "sourceRoot": "libs/dispatch/api-models/src",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "lint": {
+          "builder": "@nrwl/linter:lint",
+          "options": {
```
bilalshaikh42: Change lint runner

biosimulations/libs/dispatch/api-models/src/lib/common/dispatch-simulation.model.ts
```diff
@@ -0,0 +1,59 @@
+import { prop } from '@typegoose/typegoose';
```
bilalshaikh42, line 1: @gmarupilla Is this needed in this library, or should the dispatch/datamodel one be used?

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -130,15 +140,40 @@ export class AppController {
...
           }
         }
-      })
-
-      
+      });
     }
   }
 
+  @MessagePattern('dispatch_log')
```
bilalshaikh42: When is the "dispatch_log" message sent? Should this not be listening for the normal dispatch message?

bilalshaikh42 (changes requested):

biosimulations/nx.json
gmarupilla: Will change in upcoming commit

biosimulations/angular.json
gmarupilla: done

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: Changed the message patterns to be more self-explanatory

biosimulations/libs/dispatch/api-models/src/lib/common/dispatch-simulation.model.ts
gmarupilla: removed `dispatch/datamodel`

biosimulations/angular.json
```diff
@@ -1433,6 +1433,33 @@
...
         }
       }
+    },
+    "messages-dispatch": {
+      "root": "libs/messages/dispatch",
+      "sourceRoot": "libs/messages/dispatch/src",
+      "projectType": "library",
+      "schematics": {},
+      "architect": {
+        "lint": {
```
bilalshaikh42, line 1443: Change linter here as well.

biosimulations/nx.json
```diff
@@ -131,11 +131,14 @@
...
-      "tags": ["scope:dispatch", "type:datamodel", "platform:any"]
-    },
     "shared-datamodel-api": {
       "tags": ["scope:shared", "type:datamodel", "platform:any"]
+    },
+    "dispatch-api-models": {
+      "tags": ["scope:dispatch", "type:datamodel", "platform:any"]
+    },
+    "messages-dispatch": {
+      "tags": ["scope:shared", "type:messagequeues", "platform:any"]
```
bilalshaikh42: If the scope is intended to be shared. lets make this just messages. not message-dispatch. 
The type should be datamodel
The platform should be server

biosimulations/apps/dispatch-service/src/app/app.controller.spec.ts
```diff
@@ -4,11 +4,13 @@ import { ConfigService } from '@nestjs/config';
 import { HpcService } from './services/hpc/hpc.service';
 import { SshService } from './services/ssh/ssh.service';
 import { SbatchService } from './services/sbatch/sbatch.service';
+import { SchedulerRegistry } from '@nestjs/schedule';
 import {
   ClientProxyFactory,
   Transport,
   NatsOptions,
 } from '@nestjs/microservices';
+import { exception } from 'console';
```
bilalshaikh42: Should this be here?

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -80,14 +90,14 @@ export class AppController {
   }
 
   // TODO: Add API to send required info dispatch_finish pattern to NATS
-  @MessagePattern('dispatch_finish')
-  async dispatchLog(data: any) {
+  @MessagePattern(MQDispatch.SIM_HPC_FINISH)
+  async dispatchFinish(uuid: string) {
```
bilalshaikh42, line 94: Im still not understanding here. It seems that the monitoring cron job is not started until the HPC_Finish message is received. Should it not be when the job is started?

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -123,22 +134,45 @@ export class AppController {
...
           }
         }
-      })
-
-      
+      });
     }
   }
 
+  @MessagePattern(MQDispatch.SIM_DISPATCH_FINISH)
```
bilalshaikh42, line 152: Here as well, im confused as to why the finish is when the job is started? 
Does finish refer to the process of dispatching?

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: The executes when an actual simulation finishes on HPC

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: This message means, execute the below function as soon as the process of dispatching the simulation is finished.
Because it seems more intuitive to write it in this way where we actually emit the message.

biosimulations/angular.json
gmarupilla: done

biosimulations/nx.json
gmarupilla: done

biosimulations/apps/dispatch-service/src/app/app.controller.spec.ts
gmarupilla: removed

bilalshaikh42 (approved):

PR #1173, opened 2020-09-10
Updated documentation per renaming of GitHub, Docker Hub organizations and repos

jonrkarr:
Updating URLs and email addresses for renaming of GitHub, Docker Hub organizations and repositories. E.g.,
- github.com/reproducible-biomedical-modeling --> biosimulations, biosimulators
- dockerhub/crbm --> biosimulators


biosimulations/apps/dispatch/src/app/components/help/about/about.component.html
```diff
@@ -256,8 +256,8 @@
     </biosimulations-text-page-content-section>
```
bilalshaikh42, line 256: Should we distinguish the teams for each app separately? I thought that we were going to use Biosimulations team for all the apps?

biosimulations/apps/simulators/src/app/help/about/about.component.ts
```diff
@@ -9,7 +9,7 @@ export class AboutComponent implements OnInit {
   // TODO: get from app config
   apiUrl = 'https://api.biosimulations.org/'
```
bilalshaikh42, line 11: Here as well, I think it would make sense to portray each app as built and managed by the biosimulations team, with a common email/author

biosimulations/apps/dispatch/src/app/components/help/about/about.component.html
jonrkarr: Let's distinguish separate teams for BioSimulations and BioSimulators. Right now, they are the same people. That won't necessarily be the case. More importantly, I think its clearer for users to say that BioSimulations was developed by the BioSimulations team and BioSimulators was developed by the BioSimulators team.

biosimulations/apps/simulators/src/app/help/about/about.component.ts
jonrkarr: We need to have separate team names and contact information. This is important for distinguishing these as separate resources. Its also important for enabling others to get involved in specific parts of the overall ecosystem without having to take on the whole thing. Going forward, the teams won't necessarily be identical. Arguably, there are already several more people tangentially involved with BioSimulators than BioSimulations -- James Feader, Ali Sinan Saglam, Chris Myers, Thomas Payne, Brian Drawert -- due to the need for help creating containers. We should do more to credit them on the website.

For example, many of the COMBINE projects (SBML, SED-ML, KISAO, SBO, COMBINE) are led by the same people. But because the teams don't completely overlap, its important to distinguish different teams.

jonrkarr (commented):
Clarified reason for separate teams and contact information. I think this is important. Although the teams are largely the same now, this won't necessarily be the case going forward. Also, sending BioSimulators help to info@biosimulations.org may be confusing. Forwarding GitHub issues for BioSimulators to BioSimulations is also confusing, but I think developers who use issues will be more likely to understand the reason for this.

biosimulations/apps/dispatch/src/app/components/help/about/about.component.html
bilalshaikh42: That makes sense definitely.

bilalshaikh42 (approved):

biosimulations/apps/dispatch/src/app/components/help/about/about.component.html
bilalshaikh42: Should we also do the same for the runBioSimulations? That at least seems to be more of a subset of BioSimulations

PR #1228, opened 2020-09-22
Integrate dispatch frontend and API for end-to-end simulation visualization

gmarupilla:
Simulation dispatch working frontend


gmarupilla:
@bilalshaikh42 I had to revert changes from #1229 because that PR only removed the prefix from main.ts but not from other places it was being used, all the dispatch apps and proxies were breaking due to that.

biosimulations/apps/dispatch-api/src/main.ts
```diff
@@ -12,12 +12,14 @@ import { CustomOrigin } from '@nestjs/common/interfaces/external/cors-options.in
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
+  const globalPrefix = 'api';
```
bilalshaikh42: I dont think we want to add this back, correct?

bilalshaikh42 (changes requested):

bilalshaikh42:
> @bilalshaikh42 I had to revert changes from #1229 because that PR only removed the prefix from main.ts but not from other places it was being used, all the dispatch apps and proxies were breaking due to that.

Let's remove the prefix. The only place that should have the url hardcoded is the URL library correct? 

The library can be updated as well.

gmarupilla:
Removing 'api' as prefix was breaking dispatch frontend,Since the proxy expects 'api' and also all the urls in libs/urls. Ts still have 'api'.. If we are removing from one place, we need to remove it from all the places and fix proxy as well

biosimulations/apps/dispatch-api/src/main.ts
gmarupilla: removed

biosimulations/apps/dispatch-api/src/app/app.controller.ts
```diff
@@ -105,7 +104,7 @@ export class AppController implements OnApplicationBootstrap {
     const simSpec: SimulationDispatchSpec = {
       authorEmail: bodyData.authorEmail,
       nameOfSimulation: bodyData.nameOfSimulation,
-      simulator: bodyData.simulator,
+      simulator: bodyData.simulator.toUpperCase(),
```
bilalshaikh42: The simulators at [biosimulators](http://biosimulators.org/simulators) will have all lowercase id's. We should be consistent between the two

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-api/src/app/app.controller.ts
bilalshaikh42: Just want to make sure this won't break anything on the backend. If not, I can make the change and then merge it in

gmarupilla:
No, it’s not going to break anything at least for now. Eventually, I’ll plug-in simulator-api.

bilalshaikh42 (approved):

PR #1231, opened 2020-09-22
working on parameterzing allowed hosts to environment (product, dev, …

jonrkarr:
…test); setting login/logout redirect URLs based on environment


bilalshaikh42:
To confirm, running development locally will not be possible against the current production api. Is there a need for this?

jonrkarr:
No, I don't need to run the API locally.

My changes only partially address this todo. Until their are multiple deployments of the APIs, the issue can't be fully addressed.

jonrkarr:
@bilalshaikh42 this is ready for merging

bilalshaikh42:
This will be preventing local  development frontends from querying the api. I don't see a reason to do that. I did overlook the redirect URI however. I can include that change in the next push

jonrkarr:
Access from localhost is still allowed. Eventually this access should be removed. But this can't be removed yet.

This also fixes an issue with the login/logout redirect URLs

PR #1232, opened 2020-09-22
fixing toc

### LINKED ISSUE #1230: Table of contents does not update when used within a components with onPush strategy
**Please describe the bug**

When using a responsive table, the table of contents are not generated at the same breakpoint

**Please describe how to reproduce the bug**

Go to biosimulators.org/copasi/

1. resize the window until the mobile table appears
2. contents will not be generated
3. continue to shrink the window, contents will appear

**How should BioSimulations have behaved**

Contents should be generated at the same breakpoint as the switch to mobile table

jonrkarr:
fixes #1230


bilalshaikh42:
I am not sure if this is the best way to approach this. While this works for this component, it means that no page that is using the UI components can use the onPush strategy, which will cause a large performance hit, especially for tables.  We should try to find the UI component that is actually failing to update, and try to fix the problem there.
I made this change in #1233 anyway, but will reopen the issue to try and fix

PR #1235, opened 2020-09-22
Change detection

### LINKED ISSUE #1230: Table of contents does not update when used within a components with onPush strategy

bilalshaikh42:
Changes the toc directive to trigger change detection 

closes #1230


bilalshaikh42:
It seems this too is a partial solution.  While it works for page resizes, it does not work for the initial load. This is because the page resize is an event that I have been able to set to trigger change detection. I believe for the initial load the only option is to use the default change detection strategy.

PR #1280, opened 2020-09-30
Remove six-month-old simulation data and add rimraf-based cleanup job

### LINKED ISSUE #1157: Setup chron job to delete dispatch simulation results after 6 months

gmarupilla:
- [x] Fixes #1157


gmarupilla:
@bilalshaikh42 Can you have look and merge my PR?

bilalshaikh42:
This will have to wait until after Combine.

A few thoughts- The api-models library should not have the file methods.
Api Models should just be the return types of the api. The front end will
be loading the API-Models library, and if it has file management
dependencies (like fs and rimraf) that will cause it to either fail or load
a lot of extra code to the browser.

Secondly, if the database has the information about when the models are
submitted I'm not sure what the purpose of the cron job is. This can be
handled with a simple bash script at some point to get a list of uuids to
delete if we need to. The cron job is adding a lot of code to the service
that is really not its main purpose.

gmarupilla:
1. Can I move one level up to  `@Biosimulations/dispatch/file-modifiers`?

2. What do you propose to do for bash script, should it be executed from NestJS code? Or should it be executed in a cron from a server? Or something else?

bilalshaikh42:
@gmarupilla can some of these todos be resolved?

gmarupilla:
> @gmarupilla can some of these todos be resolved?

Yes I can do.

gmarupilla:
> > @gmarupilla can some of these todos be resolved?
> 
> Yes I can do.

@bilalshaikh42 one todo remaining from this PR, you can merge

bilalshaikh42 (approved):

PR #1375, opened 2020-10-21
Add validation route to SImulators API

bilalshaikh42:


jonrkarr (approved):
I think this looks good. I see there's now a `validate` endpoint which doesn't require authentication and returns 204 or 400. Should this endpoint will be easy to incorporate into the validation GitHub action workflow.

Issue #1391, opened 2020-10-23
Stricter simulator specifications validation


jonrkarr:
Currently the validation seems to allow extra keys at any point in the schema. While this flexibility may be helpful for us as developers, I think this could give the community the false impression that BioSimulators can use such keys. 

- [x] We need to make the validation stricter. The one part of the schema would come remain more flexible is the `biosimulators` field. I believe this is addressed in #1421.
- [x] ~~If you want to allow the community to contribute additional information that is opaque to the BioSimulators schema/tools, there could be a specific field for such information e.g., `annotation`. This field should be optional and the type of its value should be any. This is roughly how SBML handles extra user-specific data. I don't think we have a need for this now.~~
- [x] ~~the update endpoint (PUT `/simulators/{}/{}`) currently ignores the simulator and version parameters. These should be removed or validated that they match the id and version in the specification (body of the PUT). This is ok as is. As is, this allows renaming of simulators or their versions.~~

bilalshaikh42:
done in #1435

Issue #1401, opened 2020-10-28
Return biosimulators metadata


jonrkarr:
- [x] I think `schemaVersion` could be unclear. Rename to `specificationsVersion`. -- done in #1421.
- [x] We need to have the ability to return the biosimulators metadata. I think this may also be fixed by #1421.
  - `schemaVersion`
  - `imageVersion`
- [x] This part of the schema should appear in the documentation at api.biosimulators.org. I think this may also be fixed by #1421.

bilalshaikh42:
The biosiumulators metadata should return. The current database simply does not contain entries with the metdata defined. Once the entries are updated, the value will return

bilalshaikh42:
For example, this call returns the data https://api.biosimulators.org/simulators/latest?id=gillespy2

jonrkarr:
Looks good, except I'd like to rename `schemaVersion` to `specificationsVersion` to be clearer about the meaning of this field. I changed this in #1421.

PR #1421, opened 2020-11-03
Stricter validation of simulators

### LINKED ISSUE #1398: Validate that all ontology terms defined in specifications exist (key that they are defined in the ontology)
Example: KISAO_0000036 at https://biosimulators.org/simulators/copasi/latest

- KISAO_0000036 passes the validation in the BioSimulators API
- But the UI can't find KISAO_0000036

### LINKED ISSUE #1400: Validate values of enumerations in simulator specifications
E.g. validate that the value is a valid enumeration of AlgorithmParameter.type

### LINKED ISSUE #1417: Add additional enumerations to AlgorithmParameter.type for KiSAO ids and any
The current types are `boolean`, `number`, and `string`. 

We need to add a couple of types
- `kisaoId`: needed to capture nested algorithms. e.g., an option of the CMBPy FBA algorithm is the simplex method (KISAO_0000550) 
- `any`: needed to capture values that are more complex datatypes such as a list of strings (e.g., list of reactions to calculate bounds for for flux variability analysis) or a dictionary (e.g., weighted probabilities of transitions in asynchronous logical simulation)

### LINKED ISSUE #1418: Update simulator parameters table to support parameters with type=kisaoId
https://github.com/biosimulations/Biosimulations/blob/d7d57252318d6b3b504a21f661ff4c2b09d5fee4/biosimulations/apps/simulators/src/app/simulators/view-simulator/view-simulator.service.ts#L142-L147

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in d7d57252318d6b3b504a21f661ff4c2b09d5fee4. It&#x27;s been assigned to @jonrkarr because they committed the code.

jonrkarr:
- Adding additional enumerations for AlgorithmParameter.type; closes #1417, closes #1418 
- Renaming `schemaVersion` to `specificationsVersion`; addresses #1401.
- Annotating type of `biosimulators` attribute. I believe this addresses the rest of #1401.
- Validating that EDAM, KiSAO, SBO, SPDX ids exist; closes #1398
  Note, this requires changing the scope of `ontology-sources`. This causes a circular dependency which needs to be resolved.
- Validating enumerations; closes #1400
- Throw errors on extra fields; I think this addresses #1421. @bilalshaikh42 please check. I can't tell from the Mongoose documentation if [`useNestedStrict`](https://mongoosejs.com/docs/guide.html#useNestedStrict) also needs to be set.

### DIFF
--- a/biosimulations/apps/dispatch-api/src/app/app.module.ts
+++ b/biosimulations/apps/dispatch-api/src/app/app.module.ts
@@ -13,6 +13,9 @@ import { ScheduleModule } from '@nestjs/schedule';
 import { ModelsModule } from './resources/models/models.module';
 import { AppService } from './app.service';
 
+import * as mongoose from 'mongoose';
+mongoose.set('strict', 'throw');
+
 @Module({
   imports: [
     BiosimulationsConfigModule,
--- a/biosimulations/apps/dispatch-service/src/app/app.module.ts
+++ b/biosimulations/apps/dispatch-service/src/app/app.module.ts
@@ -17,6 +17,9 @@ import { MongooseModule } from '@nestjs/mongoose';
 import { ModelsModule } from './resources/models/models.module';
 import { SimulationService } from './services/simulation/simulation.service';
 
+import * as mongoose from 'mongoose';
+mongoose.set('strict', 'throw');
+
 @Module({
   imports: [
     BiosimulationsConfigModule,
--- a/biosimulations/apps/platform-api/src/app/app.module.ts
+++ b/biosimulations/apps/platform-api/src/app/app.module.ts
@@ -11,6 +11,9 @@ import { BiosimulationsAuthModule } from '@biosimulations/auth/nest';
 import { TypegooseModule } from 'nestjs-typegoose';
 import { MongooseModule } from '@nestjs/mongoose';
 
+import * as mongoose from 'mongoose';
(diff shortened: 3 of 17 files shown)


bilalshaikh42:
I think it would be better to import the Ontology service and use that for the verification. I can push this shortly.

bilalshaikh42:
The strict option needs to be defined in the schema. We do not want to import the mongose module directly as it is imported through the framework.

jonrkarr:
Others recommend importing mongoose directly even though it is imported through MongooseModule. The is the only way I can find to set global options.

bilalshaikh42:
> Others recommend importing mongoose directly even though it is imported through MongooseModule. The is the only way I can find to set global options.

I don't think that setting this globally is what we want. This would prevent things such as logging and having unstructured data for some objects. This can be defined on a per schema bases as in #1424. Each Module for each datatype will have its own mongoose object with distinct options.

jonrkarr:
It's fine to set strict validation this way (locally on every schema). My understanding from the documentation is that this option has to be set on every schema.

jonrkarr:
This pull request does more than add strict validation.

bilalshaikh42:
> This pull request does more than add strict validation.

Sorry, meant to close #1391 only. Not sure why it closed the PR

PR #1437, opened 2020-11-05
Delete duplicate-case mongo filter folder to fix case sensitivity

gmarupilla:
The PR deletes an empty file named `mongo`. On windows and mac, the same name with different cases (a file named `mongo` and a folder named `Mongo`) gives problems on git. The paths are not case sensitive on Windows and Mac git.
I rarely use Linux due to performance issues. So I thought it's good to delete this file since it's empty anyway.


bilalshaikh42 (approved):
Sorry, my mistake. Did not realize this file was committed

PR #1485, opened 2020-11-10
Add MQ queue config and remove unused DB connection in dispatch (closed)

### LINKED ISSUE #1481: Dispatch API/Service do not use message queue
The dispatch client and API are sending messages without a queue. A queue should be used to ensure that only one simulation is run for each submission. The service will run multiple simulations if it is scaled to more than 1 replica.

### LINKED ISSUE #1483: Dispatch API/Service are using both Mongoose and Typegoose
Only one of these packages should be used.
@gmarupilla  Which package are you using for the database connection?

gmarupilla:
fixes #1481 
fixes #1483


bilalshaikh42:
I have used the mongoose database connection as a part of #1484. Lets close this in favor of that one. 
I will work on saving the run information. The omex files can also be stored in the mongodb for now

PR #1520, opened 2020-11-16
Add simulation-run resource with API, message queue, and HPC sbatch wiring

### LINKED ISSUE #1482: Dispatch API/Service share Database connection
Both the dispatch and the service are connecting to the database. This is probably not the best design option. Communication between the API and service should be through messages, not the database.

gmarupilla:
@bilalshaikh42 /run is now working with an actual simulation run from HPC.


biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.controller.ts
```diff
@@ -123,6 +129,20 @@ export class SimulationRunController {
...
+    // console.log('start to Dispatch with filename: ', file.originalname)
+    this.messageClient.send(MQDispatch.SIM_DISPATCH_START,
+      {
+        simulationId: run.id,
+        omexFileName: file.originalname,
+        simulator: run.simulator,
+        simulatorVersion: run.simulatorVersion
+      })
+      .subscribe(dat => {
+        // TODO: Make MQ work without this subscription
```
bilalshaikh42: If you make this into emit rather than send it will not need a subscription. However, We do want this to be a message because we want the dispatch service to return an acknowledgment that the run was accepted.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -52,20 +55,22 @@ export class AppController {
...
     const simulatorString = `biosimulations_${data.simulator}_${data.simulatorVersion}`;
-    const hpcTempDirPath = `${this.configService.get('hpc.hpcBaseDir')}/${data.uniqueFilename.split('.')[0]
-      }`;
+    const hpcTempDirPath = `${this.configService.get('hpc.hpcBaseDir')}/${data.simulationId}`;
     const sbatchString = this.sbatchService.generateSbatch(
       hpcTempDirPath,
       simulatorString,
-      data.filename
+      data.omexFileName,
+      // 'http://99d1dcb20ebd.ngrok.io',
```
bilalshaikh42: Lets get this out of the codebase

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -74,8 +79,8 @@ export class AppController {
     this.hpcService.dispatchJob(
       hpcTempDirPath,
       sbatchPath,
-      omexPath,
-      data.filename
+      // omexPath,
+      data.omexFileName
     );
 
     return { message: 'Simulation dispatch started.' };
```
bilalshaikh42: This is the message that is going to be sent back. So the above api should ensure that it gets this message and handles the error if not. 
The message should probably be changed to something more informative like the time or other info about the status

biosimulations/apps/dispatch-service/src/app/services/hpc/hpc.service.ts
```diff
@@ -33,18 +33,18 @@ export class HpcService {
           'Simdirectory created on HPC: ' + JSON.stringify(value)
         );
 
-        this.sshService
-          .putFile(omexPath, `${simDirBase}/in/${omexName}`)
-          .then((val) => {
-            this.logger.log(
```
bilalshaikh42, line 39: Lets remove instead of commenting out. We have the git history if it is needed

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -1,3 +1,4 @@
+import { urls } from './../../../../libs/config/common/src/lib/urls';
```
bilalshaikh42: Cant use this relative import. Needs to be changed to the @biosimulations/ pattern

bilalshaikh42 (changes requested):
Looks great! 
Using wget on the sbatch template seems to have simplified things definitely. Let's also determine where the service will use the patch method of the api to update the status when the job is created.
A few changes are needed, if you can work on those. 
I will also work on adding schema and typing for the message payloads. I'll try to get this merged today pending #1521.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -52,20 +55,22 @@ export class AppController {
...
 
-    const omexPath = data.filepathOnDataStore;
-    const sbatchName = `${uuid()}.sbatch`;
+    const sbatchName = `${data.simulationId}.sbatch`;
     const sbatchPath = path.join(sbatchStorage, sbatchName);
 
     this.logger.log('SBatch path: ' + sbatchPath);
 
     // Generate SBATCH script
+    // TODO: Rename singularity images biosimulations_ to biosimulators_ on HPC and build new images according to /simulator from DB
```
bilalshaikh42: Should we use the API here to get the name of the docker image? The singularity images can then be called the same

biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.controller.ts
```diff
@@ -1,3 +1,4 @@
+import { MQDispatch } from './../../../../libs/messages/src/lib/messages';
```
bilalshaikh42: This import also needs to be fixed

biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.module.ts
```diff
@@ -28,7 +34,21 @@ import { SimulationRunService } from './simulation-run.service';
         schema: SimulationFileSchema,
       },
     ]),
+    
+  ],
+  providers: [
+    SimulationRunService,
+    {
+      provide: 'DISPATCH_MQ',
```
bilalshaikh42, line 42: This can be separated into it own module. 
@gmarupilla  don't worry about addressing this, I can work on it.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
```diff
@@ -52,20 +55,22 @@ export class AppController {
       return { message: 'Unsupported simulator was provided!' };
```
bilalshaikh42: Need to  define messages via enum or schema

bilalshaikh42 (changes requested):

biosimulations/apps/dispatch-service/src/app/app.controller.ts
bilalshaikh42: Same for this one, I will work on created well defined messages

biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.controller.ts
gmarupilla: Will change this as well.

biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.controller.ts
gmarupilla: Yes, I checked with emit, but the problem persists the same.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: Sure, will do it.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: While testing in my local, the URL from libs appending `//api` instead `localhost:3333/api`. So, I'm serving it through `ngrok` to download the omex file on HPC(for local run only). We might need to change the local config in the URL file in libs.

biosimulations/apps/dispatch-service/src/app/services/hpc/hpc.service.ts
gmarupilla: Sure, will do it.

gmarupilla (commented):
> Looks great!
> Using wget on the sbatch template seems to have simplified things definitely. Let's also determine where the service will use the patch method of the api to update the status when the job is created.
> A few changes are needed, if you can work on those.
> I will also work on adding schema and typing for the message payloads. I'll try to get this merged today pending #1521.

Sure, I'll push the requested changes

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: I'll remove

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: Yes, we can use the api to get the name of the docker images.

biosimulations/apps/dispatch-service/src/app/app.controller.ts
gmarupilla: Resolving this because the code is updated.

PR #1575, opened 2020-11-29
Update ci2.yml

### LINKED ISSUE #1574: Docker images are generated even if tests fail
If the tests fail for any library or app, docker containers should not be generated. This will cause inconsistent deployments across apps  and some apps to have bugs that make it to the images.

bilalshaikh42:
closes #1574


bilalshaikh42:
Merged into dev

PR #1587, opened 2020-11-30
Ci

bilalshaikh42:


jonrkarr (dismissed):
It would be helpful to retain linting and testing for all branches, but I guess the same thing can be achieved with draft pull requests.

PR #1628, opened 2020-12-03
Typing

bilalshaikh42:


jonrkarr (approved):
I looked at `biosimulations/libs/datamodel/common/src/lib/utils.ts `
*  `getSchemaUserPaths` should continue to use `Schema.paths`.
* `next(undefined)` seems fine in `addValidationForNullableAttributes`.

I reverted the first change and pushed it to this branch. With this change, the validation appears to be working correctly.

bilalshaikh42:
It looks like this was indeed a bug in mongoose, so revert makes sense. I'll hold on this for further updates 

Automattic/mongoose/issues/9642

jonrkarr:
I only reverted the change from `Schema.paths` to `Schema.path`. The other changes look fine.

PR #1703, opened 2020-12-11
aligning simulation status enum data models

jonrkarr:


biosimulations/libs/datamodel/common/src/lib/resources/primary.ts
```diff
@@ -5,7 +5,6 @@ import {
   BiomodelAttributes,
```
bilalshaikh42, line 5: I would rather keep the dispath api out of the platform app for now. Since we dont really have a working platform app, we can decide later the best way to include this. But the dispatch/api-models should be scoped only to the dispatch front end and api. 
Later, we can make a library that handles submitted runs

bilalshaikh42 (changes requested):

PR #1739, opened 2020-12-15
implement getting reports

bilalshaikh42:


bilalshaikh42:
> ## The api does not keep track of higher level sedml file
> https://github.com/biosimulations/Biosimulations/blob/652e29eb371664493c7041988518df126762396c/biosimulations/apps/dispatch/src/app/services/visualization/visualization.service.ts#L21-L26
> 
> ###### This comment was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in [652e29e](https://github.com/biosimulations/Biosimulations/commit/652e29eb371664493c7041988518df126762396c) in #1739. cc @biosimulations.

@jonrkarr  Here is the issue.

PR #1805, opened 2020-12-21
adding examples and documentation

jonrkarr:
- adding examples for GillesPy2, BioNetGen; adding example for COPASI
- adding documentation about SED-ML execution


jonrkarr:
Error due to nrwl installation

PR #1853, opened 2020-12-30
Deploy new changes

jonrkarr:
@bilalshaikh42 
- Before this is merged, these new images need to be pulled into the cluster
  - `ghcr.io/biosimulators/pysces:0.9.9`
  - `ghcr.io/biosimulators/cobrapy:0.20.0`
  - `ghcr.io/biosimulators/cbmpy:0.7.25`


bilalshaikh42:
I have added in the new images. Before merging, we need to manually check that the dev deployment is working end to end. 
At some point, we will need to add in some automated tests for that.

bilalshaikh42 (dismissed):

jonrkarr:
@bilalshaikh42 Yes, we should manually check changes that affect runBioSimulations while we're waiting on review from NAR. 

At the moment, I just want to deploy `simulators-api` and `simulators`. biosimulators.dev looks good. 

From what I can tell, merging this PR will trigger the deployment repo to deploy new versions of all of the apps. How should I deploy just the 2 apps?

bilalshaikh42:
The repo is designed so that all of the applications always remain in sync. This way a change to the data structures of one app won't cause other depending app to fail since that would be enforced at build time. 

For now, we could just manually change the tag of the image in the deployment repo to the latest Dev tag. But overall, it's really not designed to work that way.

We could look into allowing for more independent deployments, but a better solution would be better testing

bilalshaikh42 (dismissed):

jonrkarr:
Deploying the apps in sync makes sense. That's what's being tested. This policy works well for the apps in this repository, but doesn't quite cover the entire ecosystem.

These changes aren't ready to merge. I noticed at least 1 issue with the example simulations. See #1863.

I can work around this for now by using an environment variable to direct the BioSimulators test suite to work with `api.biosimulators.dev` instead `api.biosimulatos.org`.

bilalshaikh42 (approved):

Issue #1881, opened 2021-01-05
Dispatch app is polling simulation runs that have been completed


bilalshaikh42:
On run.biosimulations.org, loading the example simulations causes the app to poll for their status, even though all are marked as completed.

jonrkarr:
This might be related to the local storage issue #1882. If the state can't be saved, it could keep polling.

jonrkarr:
The status will be polled while either of these conditions are true:
- The status of a simulation is `CREATED`, `QUEUED`, or `RUNNING`
- The status of a simulation `SUCCEEDED` and `resultsSize` is `undefined`

(a) The `resultsSize` attribute is undefined for the previously run "example" simulations. This will cause the polling for status to repeat forever. The "example" simulations need to be re-run, which I did in #1894 . @bilalshaikh42 do you see this problem when "your simulations" only includes simulations that you've run recently? My browser doesn't keep polling for status if I only have newer simulations.

(b) The reason that the polling checks for `resultsSize` is that it needs `resultsSize` to determine whether to display a download button. It look like the `resultsSize` is updated after status is set to SUCCEEDED. @bilalshaikh42 is that correct?

jonrkarr:
@bilalshaikh42 do you still see this issue?

bilalshaikh42:
Nope, seems to be resolved

Issue #1797, opened 2020-12-20
Some successful simulation runs do not show log or results


moraru:
single sim run with tellurium succeeds, but two links seem broken

the job is:
```json
  {
    "id": "5fdf1c4ff10d68cf08393b54",
    "name": "tl-vilar1",
    "email": null,
    "simulator": "tellurium",
    "simulatorVersion": "2.1.6",
    "submittedLocally": true,
    "status": "SUCCEEDED",
    "runtime": 8501,
    "submitted": "2020-12-20T09:41:35.335Z",
    "updated": "2020-12-20T09:41:43.844Z",
    "resultsSize": 28050,
    "projectSize": 9058
  }
```

clicking on either log or viz produces error 500 empty page

clicking on download works, returning a zip file containing logs and a pdf file of results report which looks ok

jonrkarr:
I suspect this is related to the simulation service, rather than the tellurium image. The tellurium image has no role in whether logs are saved, retrieved, or displayed.

moraru:
i have same problem running with vcell image - it runs succesfully, download button works, viz and log links show error 500

https://run.biosimulations.org/simulations/5fdf08f0391352059f26ecb9

jonrkarr:
In this instance, the problem is because the results API returns a 404 error
https://run.api.biosimulations.org/results/5fdf08f0391352059f26ecb9?sparse=true

The UI is just passing along this error to the user

moraru:
> In this instance, the problem is because the results API returns a 404 error
> https://run.api.biosimulations.org/results/5fdf08f0391352059f26ecb9?sparse=true
> 
> The UI is just passing along this error to the user

but why is the download button working?  are there different API calls to fetch the results, some which work and some which do not?

jonrkarr:
They're different endpoints. 
- 1 endpoint retrieves a zip archive of CSV files which is produced when the simulation results are first saved. 
- A second endpoint retrieves the results as a JSON data structure.

moraru:
It actually is specific to tellurium, and it actually depends on the omex file.
This happens when tellurium produced PDF instead of CSV.
The API is correct as it is being asked for CSV but there is none.

jonrkarr:
To help us debug, please post the OMEX files associated with each error.

Whether or not a CSV file is produced depends on whether a `report` is defined in the SED-ML file. tellurium not generating a CSV file is not a tellurium error.

[...]

Issue #1851, opened 2020-12-29
Read simulation status output into simulations database


jonrkarr:
The simulator containers built with BioSimulators-utils (AMICI, COPASI, GillesPy2, PySCeS and soon BoolNet, BioNetGen, CBMPy, COBRApy) will save status information when they are next rebuilt.

These containers will also now produce reports (and eventually plots) progressively. Meaning, 
- Reports will be initialized once the first task completes. 
- Once at least one data set of a report has been generated, the HDF5 and CSV outputs will contain the data sets that have been generated
- All data sets which haven't been generated yet will have the value `NaN`
- Reports for which no data sets have been generated will have no columns
- Plots will behave similarly.
- Note: For now, reports are initialized with rows, but no other dimensions. The shape of the other dimensions is not realized until the first data has been produced. Its implemented this way because each KISAO term can produce a different shape of output (i.e., columns for non-spatial algorithms or multiple dimensions for spatial algorithms). We need the individual simulation tools to tell us this information about the other dimensions. The simulations are not yet passing such information to the initialization of the reports.

## Location
`{ out-directory }/status.yml`

### Rationale
I felt that status information should be saved to a separate file rather than entangled with reports (and HDF5) as we discussed for a few reasons
- Not all COMBINE archives generate reports. In fact, archives aren't required to have any outputs. Thus, we need to separate status and reports a bit
- It seemed to me that status files should be light-weight
- It seemed to me that status should either use a log-like format or something that can easily accommodate a straight-forward data structure. HDF5 doesn't work well for either.
- I chose YAML because its easily human-readable

### Alternatives
Obviously, YAML could easily be changed to JSON, which is slightly easier for use with TypeScript. But, I assume there's a YAML TypeScript package we can use to read these files.

### Data model
See https://github.com/biosimulators/Biosimulators_utils/blob/dev/biosimulators_utils/exec_status/data_model.py

### Example (in JSON format)
The potential status values are `QUEUED`, `RUNNING`, `SUCCEEDED`, `SKIPPED`, and `FAILED`.

```json
{
    "status": "SUCCEEDED",
    "tasks": {
        "task_1_ss": {"status": "SUCCEEDED"},
        "task_2_time_course": {"status": "SUCCEEDED"},
    },
    "outputs": {
        "report_1": {
            "status": "SUCCEEDED",
            "dataSets": {
                "dataset_1": "SUCCEEDED",
                "dataset_2": "SUCCEEDED",
[...]

PR #1931, opened 2021-01-12
Simulation status

bilalshaikh42:
I have added a intermediate status for simulation runs called "processing".

This is intended to indicate that the job has finished on the HPC and will the results will be available soon. This was previously indicated as "Completed". However, due to the asynchronous reading of files, the simulation run would get marked as completed before the results were uploaded.  Now, a simulation run is marked as completed only after the results have been updated. 

This should prevent some of the confusion and need to check for edge cases that was seen in #1881. This will NOT resolve the issue seen in #1797 as a run that does not produce any output will be marked as completed as soon as the dispatch service "uploads" the empty results.


bilalshaikh42:
@jonrkarr Would this change cause any issues for the way you are outputting simulation progress? 
I don't think this needs to  be reflected in #1851, as I see that as a different measure of "progress" than the status here.

If this change looks okay, it should automatically merge in on approval.

bilalshaikh42:
@jonrkarr Could you confirm that these changes are okay? This is ready to merge

jonrkarr:
This is helpful. This addresses the issue Ion and I were discussing that the STATUS was set to SUCCEEDED before the results were available.

Regarding #1851, this makes a similar distinct between the status of tasks and outputs.  After a task finishes, its status will be SUCCEEDED (or FAILED), but the status of the outputs that it contributes to will be QUEUED (or RUNNING if the output also contains results from prior tasks). By the end of the execution of the entire COMBINE archive, the status of each task and output will be SUCCEEDED (or FAILED).

I see #1851 as somewhat distinct from the status and log that's currently implemented. The current status and log cover the entire creation, saving, and execution of a COMBINE archive. #1851 captures more detail about the execution. Since #1851 doesn't cover ever place where failures can occur, we still need this status and log. In #1851, I'm trying to use consistent terminology to what you've chosen, but I don't think we need to integrate the two types of logs together.

bilalshaikh42:
> This is helpful. This addresses the issue Ion and I were discussing that the STATUS was set to SUCCEEDED before the results were available.
> 
> Regarding #1851, this makes a similar distinct between the status of tasks and outputs. After a task finishes, its status will be SUCCEEDED (or FAILED), but the status of the outputs that it contributes to will be QUEUED (or RUNNING if the output also contains results from prior tasks). By the end of the execution of the entire COMBINE archive, the status of each task and output will be SUCCEEDED (or FAILED).
> 
> I see #1851 as somewhat distinct from the status and log that's currently implemented. The current status and log cover the entire creation, saving, and execution of a COMBINE archive. #1851 captures more detail about the execution. Since #1851 doesn't cover ever place where failures can occur, we still need this status and log. In #1851, I'm trying to use consistent terminology to what you've chosen, but I don't think we need to integrate the two types of logs together.

Great, could you approve the PR? it should merge in automatically.

jonrkarr (approved):
Looks good. I see updated `apps/dispatch/src/app/services/simulation/simulation-status.service.ts` as well.

PR #1945, opened 2021-01-12
add controller for image generation

### LINKED ISSUE #1942: Image building is very slow
Building the singularity image on each run is very slow. Even though the layers are cached, the image is still rebuilt for simulators that already exist.
We can remove the singularity pull command on each run. This will lead to the problem that images will not be updated if a change is made for a new tag. A simple API endpoint should be sufficient to provide an automated way to update images

bilalshaikh42:
Provides an endpoint to trigger a refresh of the cache of the singularity images saved on the HPC.  The jobs can now be run without pulling the images.

### DIFF
--- a/biosimulations/apps/dispatch-api/src/app/app.module.ts
+++ b/biosimulations/apps/dispatch-api/src/app/app.module.ts
@@ -12,11 +12,13 @@ import {
   AuthTestModule,
   BiosimulationsAuthModule
 } from '@biosimulations/auth/nest';
+import { ImagesModule } from '../images/images.module';
 
 @Module({
   imports: [
     BiosimulationsConfigModule,
     BiosimulationsAuthModule,
+    ImagesModule,
     HttpModule,
     MongooseModule.forRootAsync({
       imports: [BiosimulationsConfigModule,],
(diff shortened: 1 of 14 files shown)


jonrkarr (approved):
This looks good.

I assume the simulation job submission waits for the image updating job to complete. I didn't read the code enough to check.

If I understand correctly, the PR doesn't determine whether an image needs to be updated. Can you use Docker to determine that outside an SBATCH job?

bilalshaikh42:
> This looks good.
> 
> I assume the simulation job submission waits for the image updating job to complete. I didn't read the code enough to check.
> 
> If I understand correctly, the PR doesn't determine whether an image needs to be updated. Can you use Docker to determine that outside an SBATCH job?

The simulations are actually completely separate from the updating. 

The jobs will use the latest image of each tag is on the HPC.  If it is a new tag, then the latest image can just be downloaded as a part of the job.  If it is an older tag that has been updated on GHCR, then it would still use the older cache. 

The endpoint that is created here would be called by the github CI or just someone manually to have the hpc update its cache for a specific tag. That way, if the image is updated, any simulations after calling this endpoint would use the up to date image

jonrkarr:
I understand the plan now.

I can incorporate a call to the new endpoint into the workflow for committing simulators into the database. If you'd like me to set that up:
- Can you describe the endpoint that should be called in biosimulators/Biosimulators_test_suite#17?
- The workflow already gets a key for the biosimulators API. If this needs a different key, can you create the key and add it as a secret for the https://github.com/biosimulators/Biosimulators repo (where the workflow runs)?

PR #2035, opened 2021-01-25
Implement getting and saving logs. Handle older simulations

### LINKED ISSUE #1978: Simulation Run "Updated" field does not change when status is not updated
When the dispatch service checks the status of a job and sends a request to the api to update it, if the status is still the same the updated field will not change. This makes sense, since there is no write to the database, and the data is the same.

However, I was thinking of the updated field from the perspective of the user/submitter. I want the updated field to reflect when the status of the simulation run was last determined. If the simulation has been running for a while, the Updated field should tell me when it was last checked and confirmed to be "running". 

The patch method of the simulation run needs to be changed to account for this.

bilalshaikh42:


bilalshaikh42:
@jonrkarr 
Based on my local testing, this seems to be working. 
The logs that are being displayed are not showing the correct formatting. I think this might be due to the limited output of the omex/simulator combination that I chose.

jonrkarr:
What does the `log.yml` file look like? Can you attach an example here? We should rule out that the simulation isn't using an older version of one of the images?

bilalshaikh42:
> What does the `log.yml` file look like? Can you attach an example here? We should rule out that the simulation isn't using an older version of one of the images?

```{
  "duration": 0.392491,
  "exception": null,
  "output": null,
  "sedDocuments": [
    {
      "duration": 0.301792,
      "exception": null,
      "location": "LSODA.sedml",
      "output": "Found 1 SED-ML documents with 1 models, 1 simulations, 1 tasks, 1 reports, and 0 plots:\r\n  LSODA.sedml:\r\n    Tasks (1):\r\n      task1\r\n    Reports (1):\r\n      report_1_task1: 2 data sets\r\n\r\nExecuting SED-ML file 0: LSODA.sedml ...\r\n",
      "outputs": null,
      "skipReason": null,
      "status": "SUCCEEDED",
      "tasks": null
    }
  ],
  "skipReason": null,
  "status": "SUCCEEDED"
}
```

This is just a random simulation that  I tested with, so it might be an older version of the logs. I had not considered that. I think the best way to test this might just be to merge in the changes to dev and run some new simulations.

jonrkarr:
Looks like you used COPASI with the Varusai example omex file. 

Below is the log you should get. This is what I generated with ghcr.io/biosimulators/copasi:4.30.233. 

It looks like the cluster is using an older image from before I changed the logging.

All of the images, except tellurium should produce logs similar to below with information about each individual task. The tellurium image produces coarser logs with information about the execution of individual SED-ML files.

```yaml
duration: 0.386083
exception: null
output: null
sedDocuments:
- duration: 0.318174
  exception: null
  location: LSODA.sedml
  output: "  Found 1 tasks and 1 outputs:\r\n    Tasks:\r\n      `task1`\r\n    Outputs:\r\
    \n      `report_1_task1`\r\n  Executing task 1: `task1`\r\n    Executing simulation\
    \ ... \e[34msucceeded\e[0m\r\n    Generating 1 outputs ...\r\n      Generating\
    \ output 1: `report_1_task1` .../home/jonrkarr/Documents/Biosimulators_utils/biosimulators_utils/warnings.py:31:\
    \ MissingReportMetadataWarning: \e[33mReports exported to {} do not contain information\
    \ about the data type or size of each data set.\e[0m\r\r\n  warnings.warn(termcolor.colored(message,\
    \ Colors.warning.value), category)\r\r\n \e[34msucceeded\e[0m\r\n\r\n  Executed\
    \ 1 tasks and 1 outputs:\r\n    Tasks:\r\n      Succeeded: 1\r\n      Skipped:\
    \ 0\r\n      Failed: 0\r\n    Outputs:\r\n      Succeeded: 1\r\n      Skipped:\
    \ 0\r\n      Failed: 0\r\n"
  outputs:
  - dataSets:
    - id: data_set_time
      status: SUCCEEDED
    - id: data_set_pmTORC1
      status: SUCCEEDED
    duration: 0.07581
    exception: null
    id: report_1_task1
    output: "/home/jonrkarr/Documents/Biosimulators_utils/biosimulators_utils/warnings.py:31:\
      \ MissingReportMetadataWarning: \e[33mReports exported to {} do not contain\
      \ information about the data type or size of each data set.\e[0m\r\n  warnings.warn(termcolor.colored(message,\
      \ Colors.warning.value), category)\r\n"
    skipReason: null
    status: SUCCEEDED
  skipReason: null
  status: SUCCEEDED
  tasks:
  - algorithm: KISAO_0000560
    duration: 0.20174
    exception: null
    id: task1
    output: ''
    simulatorDetails:
    - key: methodName
      value: deterministic
    - key: methodCode
      value: 22
    - key: parameters
      value: {}
    skipReason: null
    status: SUCCEEDED
skipReason: null
status: SUCCEEDED
```

jonrkarr:
Feel free to go ahead and merge to dev even before logging is 100% complete. The last remaining bits can be contributed in another PR.

Issue #2046, opened 2021-01-26
Log Parsing is failing on the front end 


bilalshaikh42:
**Please describe the bug**
The front end is not displaying the structured logs correctly

**Please describe how to reproduce the bug**
1. Go to https://run.biosimulations.dev/simulations/601089bcc4eb1ef2c987f3fd#tab=log

**How should BioSimulations have behaved**
Logs should be displayed with color and should show correct output. 

**Screenshots**
![Capture](https://user-images.githubusercontent.com/32490144/105909689-e47dad80-5ff5-11eb-93bc-15ff001cdbbf.PNG)

**Additional context**
Error source "valie" is undefined
```
    if ('simulatorDetails' in value && value?.simulatorDetails?.length) {
      this.formattedSimulatorDetails = value.simulatorDetails.map(
        (keyValue: SimulatorDetail): FormattedSimulatorDetail => {
          const key = keyValue.key;
          const value = keyValue.value;
          return {
            key: key,
            value: typeof value === 'object' ? JSON.stringify(value, null, 2) : value.toString()
          };
        }
      );
    }
```

bilalshaikh42:
Here is the api response that is causing the error. It seem that the simulator details contains a key for "parameters" that does not have an associated value. Since the rest of the response is valid, this must be caused by the image output. 

I can make a patch to the front end to handle this case, but it seems like the value should be reported as null

Relevant Section: 
```
           "simulatorDetails": [
            { "key": "methodName", "value": "deterministic" },
            { "key": "methodCode", "value": 22 },
            { "key": "parameters" }
          ],
```
Full Response: 
```{
  "duration": 0.37655,
  "exception": null,
  "output": null,
  "sedDocuments": [
    {
      "duration": 0.281335,
      "exception": null,
      "location": "LSODA.sedml",
[...]

PR #2049, opened 2021-01-26
hotfix for issue #2046

bilalshaikh42:


jonrkarr (dismissed):
I already pushed this exact change.

PR #2120, opened 2021-02-03
separating example simulation runs for run.biosimulations.org and run.biosimulations.dev deployments

jonrkarr:


biosimulations/apps/dispatch/src/app/components/simulations/browse/browse.component.ts
```diff
@@ -423,6 +425,10 @@ export class BrowseComponent implements OnInit {
   }
 
   loadExampleSimulations(): void {
+    const exampleSimulationsJson = environment.production
```
bilalshaikh42: environment.prod needs to be set to environment.env=="prod"

bilalshaikh42 (changes requested):

PR #2161, opened 2021-02-14
Simulators API fails for larger documents (triggered by new test cases)

### LINKED ISSUE #2160: Simulators API fails for larger documents (triggered by new test cases)
This example (111.2 kb) fails with 500 error: [specs.json.zip](https://github.com/biosimulations/Biosimulations/files/5976972/specs.json.zip)

Now that the test site has more test cases, results for 3 of 8 simulation tools couldn't be uploaded to the registry.

Temporarily, I manually worked around this by setting some of the logs for passed test cases to `''`. With that one change, the API accepts the results. Other changes that reduce the total size of the document also enable the API to accept the payload.

bilalshaikh42:
Fixes #2160


bilalshaikh42:
@jonrkarr  This should fix the issue. I am away from my dev environment at the moment so I cannot run and confirm the app is working. It's only a small change, so should not cause any issues (famous last words...).

jonrkarr:
This fixed the issue.

What is the limit? It looks like the default is 50mb, but I can't tell if there's another setting on top of this.

50mb is more than enough for the simulators API. This is 500X the largest document we're seeing now.

jonrkarr (approved):

PR #2358, opened 2021-03-30
add new datamodel

bilalshaikh42:


bilalshaikh42:
@jonrkarr  Could you take a look and test this end to end? I can't seem to get the vega rendering right, with the labels overlapping parts of the figure. I am assuming this is a resolution issue, but want to confirm that the new datamodel parsing is correct

PR #2375, opened 2021-04-02
Combine service

bilalshaikh42:


bilalshaikh42:
@jonrkarr  The implementation of the route handling is complete. The two to-dos point to where the code from biosimulations utils will go.  I need to fix some of the ci/cd before merging, but the actual code shouldn't change from here. 

The API  spec can be edited from ApiCurio and then published to a branch on the repo using the publish feature. I shouldn't be editited from the git repo. This way we don't lose any changes made in one place but not the other. The first time you may need to log in and  connect to your account/ authorize access to biosimulations. The file is published to the   ``` biosiumulations/apps/combine-service/src/spec/combine-service.yml ``` file.

jonrkarr:
Great! 

Understood that the specs should be submitted directly to the repo by ApiCurio. Its a nice app. Easy to use. I found Swagger less intuitive.

I'll work on the Python part this weekend.

jonrkarr:
I don't quite understand the connexion bit yet -- how connexion finds Python methods. I'll work on the Python side and then ask you about the connexion piece.

bilalshaikh42:
The specs contain an operation id for each path. The operation ids that I set are "handler.set_sed". Connection then looks in the handler module (handler.py) and find the method named set_sed.  It sends in the parameters to the method with the same name as defined in the schema

bilalshaikh42:
One more note about workflow. the apps/combine-service is a ```pipenv``` workspace, similar to ```npm```. From the root (apps/combine-service) 
``` pipenv install ``` will create the virtual env for you based on the ```pipfile``` and ```pipfile.lock```.   ``` pipenv install biosimulations-utls ``` will add the utils to the ```pipfile```. ``` pipenv shell ``` will activate the virtual env. This way the repo always has the same env as the deployment environment and prevents dependency version issues.

jonrkarr:
Got it. I've used pipenv a little for this same purpose.

jonrkarr:
The organization of the service looks fine.

Have you figured out how to use `@nx-python/nx-python`? This seems a bit inflexible. Not sure how to set the options appropriately. Calling flake and unittest directly pretty straightforward. Might be simpler than trying to figure out this package.

jonrkarr:
I had to manually edit (in the ApiCurio app) the specs that ApiCurio generates. When `oneOf` is used, ApiCurio adds an extra valid type (`- type: object`). This needed to be removed. Otherwise, objects could match both `object` and one of our schemas such as `SedPlot2D`.

`nx-python`'s handling of Python paths is strange. I managed to hack `nx-python` to work.
- `apps` and `apps/combineService/` have to be turned into Python modules by adding `__init__.py` files
- The tests have to import the app by file path
- The tests have to override the location of the API specifications
- The tests have to override the resolution of handlers

bilalshaikh42:
> I had to manually edit (in the ApiCurio app) the specs that ApiCurio generates. When `oneOf` is used, ApiCurio adds an extra valid type (`- type: object`). This needed to be removed. Otherwise, objects could match both `object` and one of our schemas such as `SedPlot2D`.
> 
> `nx-python`'s handling of Python paths is strange. I managed to hack `nx-python` to work.
> 
> * `apps` and `apps/combineService/` have to be turned into Python modules by adding `__init__.py` files
> * The tests have to import the app by file path
> * The tests have to override the location of the API specifications
> * The tests have to override the resolution of handlers

I  was  thinking of the best way to incorporate python into the structure we have with the minimal work.  I agree that nx-python is not fully complete, but should help avoid some of the customization needed.  

The overriding of the paths that is needed for the tests is actually a mistake on my part. In the nx workspace everything is represented as paths from the root. However, when I was developing, I was working in the package directory and evoked the program directly from python rather than using nx. The paths in the program itself should be relative to root I think.  I hardcoded these in as relative paths from the python file. 

It might also be good to move the pipenv to the root.


 . The __init__.py in the directories should not be a problem, and actually are good for having the module/packages named consistently with the nx workspace.

bilalshaikh42:
I have not done much with the CI/CD yet, but we can just hardcode in exceptions to the workflow for this app. If we end up adding more python apps later we can be tidier with handling python, but for now anything should be fine. 
On Monday, I am planning to get the `DockerfilePython` that the CI is trying to build included. Should be very straightforward. Once that is published,  just a couple of lines over at [the deployment config](https://github.com/biosimulations/Biosimulations-Deployment )will get it deployed.

jonrkarr:
I started to add handling in the CI workflow for Python.

bilalshaikh42:
@jonrkarr Could we move the bulk of the code out of the handlers.py file into their own files? Handlers.py can just call those methods as needed to keep the mapping more flexible. This would also help with splitting the handling of the HTTP with the actual business logic so we can easily change the http headers etc, for all methods at once.

jonrkarr:
The code can be divided into multiple files, e.g., with each handler defined in a separate file and imported into one module.

Alternatively, this can be managed with the operation id in the API spec. connexion/OpenAPI don't require all methods to be in a single file.

It would be helpful to decide if we're using nx-python and what directory the code is intended to be run from. This affects whether I can more easily import code from other files.

jonrkarr:
I'm trying to get the basic setup done. Getting OpenAPI/connexion/flask/nx-python/nx to play nicely has taken some trial and error. 

Once this is done, I can easily organize the handlers into separate files.

bilalshaikh42:
I think we should run from  the root biosimulation directory. I moved the pip files to the root as a part of making the dockerfile anyway.

jonrkarr:
Moving this to the root presumes that there will only be one Python service (more specifically that all Python services will share the same pipenv). That's fine with me. Setting this up takes some work. Adding more endpoints is relatively easy. Forcing all Python endpoints to use the same environment seems like a small sacrifice to make this easier to manage.

bilalshaikh42:
> Moving this to the root presumes that there will only be one Python service (more specifically that all Python services will share the same pipenv). That's fine with me. Setting this up takes some work. Adding more endpoints is relatively easy. Forcing all Python endpoints to use the same environment seems like a small sacrifice to make this easier to manage.

Yes, this is true. This is the same principle for the JavaScript apps. This will prevent having issues where we need to keep track of which python apps use which version of biosim-utils for example. If we end up with a situation where different apps do need different versions of dependencies that we cant control, then we can go back to separating them, and have the overhead of needing to manage each environment.

jonrkarr:
This should be fine for now. PyPI packages tend to be a little more specific to versions of Python and OS packages than NPM.

bilalshaikh42:
@jonrkarr This is ready to merge from my end

jonrkarr:
@bilalshaikh42 I think this is good to merge.

I organized the handlers and changed the management commands. The testing and linting is still working. I checked the Docker image also looks good.

jonrkarr (approved):

PR #2411, opened 2021-04-09
Upload created combine archive to s3

bilalshaikh42:


jonrkarr (dismissed):
Looks good!

The UI code is just about done with the two exceptions below. I just need to make sure everything tests and lints.

- Support creating COMBINE archive using models at URLs
- Support basic model changes

Issue #2385, opened 2021-04-03
Initialize plots according to SED plot specifications in COMBINE archives


jonrkarr:
- [ ] Make simulators save plot data to HDF5
- [x] Implement logic for `/combine/sedml-output-specs` endpoint with Python
- [x] Deploy endpoint
- [x] Map endpoint to `getSpecsOfSedPlotsInCombineArchive` in `visualization.service.ts`
- [x] Use `getSpecsOfSedPlotsInCombineArchive` to initialize plots in `https://run.biosimulators.org/simulations/{run-id}`

jonrkarr:
I created a skeleton of a method (`getSpecsOfSedPlotsInCombineArchive` in `visualization.service.ts`) for retrieving the specifications of the SED plots in a COMBINE archive. This method can be connected to the new Python-powered endpoint (`/combine/sedml-output-specs`).

I connected the method to the initialization of plots. I tested this with the data below with the Caravagna simulations (e.g., 6037ee366e279fedb395461c in run.biosimulations.dev).

```typescript
public getSpecsOfSedPlotsInCombineArchive(runId: string): Observable<CombineArchive> {
    return of<CombineArchive>({
      "contents": [{
        "location": {
          "path": "./BIOMD0000000912_sim.sedml",
          "value": {
            "level": 1,
            "version": 3,
            "models": [],
            "simulations": [],
            "tasks": [],
            "dataGenerators": [],
            "outputs": [{
              "_type": SedOutputType.SedPlot2D,
              "id": "plot_1",
              "name": " ",
              "curves": [{
                "id": "plot_1_T_time",
                "name": "T",
                "xDataGenerator": {
                  "_resultsDataSetId": "BIOMD0000000912_sim.sedml/BIOMD0000000912_sim/time",
                  "id": "time",
                  "name": "time",
                  "variables": [],
                  "math": "time"
                },
                "yDataGenerator": {
                  "_resultsDataSetId": "BIOMD0000000912_sim.sedml/BIOMD0000000912_sim/T",
                  "id": "T",
                  "name": "T",
                  "variables": [],
                  "math": "T"
                }
              }, {
                "id": "plot_1_E_time",
                "name": "E",
                "xDataGenerator": {
                  "_resultsDataSetId": "BIOMD0000000912_sim.sedml/BIOMD0000000912_sim/time",
                  "id": "time",
                  "name": "time",
                  "variables": [],
                  "math": "time"
                },
[...]

PR #2414, opened 2021-04-09
Create combine archive form

### LINKED ISSUE #2355: Build form for creating COMBINE archives with model files
- Select box for model file
- Select box for format of this file
- Select box for type of simulations -- uniform time course, steady state
- Input for time course parameters
  - Start time
  - Output start time
  - Output end time
  - Number of points
- Select box for algorithm (KiSAO term) -- can indicate which tools support it 
   "CVODE -- tellurium, VCell"
- Table of Input boxes for values of parameters of the implementation of the algorithm
   - Columns:
      - Name of parameter
      - Supported tools (in parenthesis default value)
      - Chosen value
      - Checkbox to include in 
   - Validate that chosen parameters are compatible with at least one tool
   - Output which tools the SED-ML will be compatible with
- Report
  - If SBML, present multiple select box to choose variables to include in report
  - If not SBML, present list of input boxes

### LINKED ISSUE #2409: Combine Endpoint has invalid example
The documentation for the /combine/create endpoint does not have a 'value' object. This makes testing difficult as the docs cannot be used to make calls without manually editing the input

### LINKED ISSUE #2412: Cannot Submit Simulations using webform
The submit button is not working on the webform.

### LINKED ISSUE #2413: Change view simulation run error handling to allow for partial failures.
Currently, any error in fetching a part of the simulation causes the entire page to be replaced with an error handler page. This means that missing logs prevent results from being visualized.  In the reverse, if a dataset is missing, the user is unable to view the logs.
The errors need to be caught at the component level rather than the app level.

jonrkarr:
- Closes #2355
- Closes #2412
  - Fixes form validation
  - Adds styling so buttons have feedback that indicates they were pushed
 - Closes #2409
   - Examples for paths and schemas in API specs
   - Adds testing to check that examples for paths are consistent with schemas
   - Note
     - This does not test the example for each schema
     - (Per OpenAPI conventions) this validation does not throw errors if examples have additional attributes that aren't part of the schema
- Closes #2413 -- error handling for individual simulation views
- Addresses part of #2385


biosimulations/apps/combine-service/src/utils.py
```diff
@@ -49,7 +49,11 @@ def save_file_to_s3_bucket(filename):
         :obj:`str`: URL for saved file
     """
     # TODO
-    return 'https://data.biosimulations.org/XYZ'
+    temp_dirname = os.path.join('/tmp', 'combine-service-static')
+    if not os.path.isdir(temp_dirname):
```
bilalshaikh42: @jonrkarr Could you take a look at the security warning and confirm that the usage is fine?

bilalshaikh42 (dismissed):

biosimulations/apps/combine-service/src/utils.py
bilalshaikh42: https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html

bilalshaikh42: https://docs.openstack.org/bandit/1.4.0/plugins/hardcoded_tmp_directory.html

jonrkarr: This temporary just so I could test its working. This is intended to be replaced by your #2411.

bilalshaikh42: Whoops, completely missed that.

PR #2428, opened 2021-04-13
fix(dispatch): Simulation results not saved for some simulations and overall status doesn't reflect such errors

### LINKED ISSUE #2416: Simulation results not saved for some simulations and overall status doesn't reflect such errors
Example (with BoolNet): https://run.api.biosimulations.dev/results/607093329bb7309709d0ecd1?sparse=true

Here's a summary of how this example is currently handled:
- There appear to have been failures both in saving results and logs.
- This example has status "succeeded" even though results and logs weren't saved.
- The results retrieval endpoints raises a 404 error, which we are interpreting to mean that the simulation didn't generate any results. 
  - The UI displays a message saying that the simulation didn't specify outputs rather than an error message that indicates that the results couldn't be retrieved.
- The log retrieval endpoint also raises a 404 error, but the meaning is clear -- the logs weren't saved.
  - The UI displays a message saying that there was an error retrieving logs.

Issues:
- The overall status doesn't account for failures in storing results and logs.
  - Partly as a result, not enough information is being collected and/or returned by the API for the UI to differentiate between (a) no results were generated and (b) there was an error in storing results. 
- It sounds like errors in saving logs may be preventing the saving of simulation results.

Resolution:
- [ ] The overall status should account for failures in storing results and logs.
    - When overall status = failed, there will be no expectation that results were saved. Thus, the UI will not try to retrieve results.
- [ ] Results should saved even if there are failures in saving logs.

bilalshaikh42:
The  simulation is now marked as failed if processing/uploading the results fails.
All results and logs are now updated independently. The issue with  results not being uploaded if logs failed was caused by the fail fast behavior of promise.All(). Using the new es2020 promise.allSetteled() fixes this problem, but requires updating the base tsconfig.

Fixes #2416

### DIFF
--- a/biosimulations/apps/dispatch-service/src/app/results/log.service.ts
+++ b/biosimulations/apps/dispatch-service/src/app/results/log.service.ts
@@ -70,6 +70,9 @@ export class LogService {
       .then((_) => {
         this.logger.debug('Sent Log to API');
       })
-      .catch((reason) => this.logger.error(reason));
+      .catch((reason) => {
+        this.logger.error(reason);
+        throw reason;
+      });
   }
 }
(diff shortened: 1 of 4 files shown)


jonrkarr (approved):
Sounds great.

PR #2538, opened 2021-05-11
Create sim s3

bilalshaikh42:


biosimulations/apps/dispatch-api/src/simulation-run/simulation-run.controller.ts
```diff
@@ -290,15 +319,19 @@ export class SimulationRunController {
...
+      'Download the COMBINE/OMEX archive file for the simulation run',
   })
   @Get(':id/download')
-  async download(@Param('id') id: string, @Res({ passthrough: true }) response: Response) {
+  public async download(
+    @Param('id') id: string,
+    @Res({ passthrough: true }) response: Response,
+  ): Promise<void> {
     const file = await this.service.download(id);
     if (file.mimetype) {
```
bilalshaikh42, line 345: Add content-size header here if available

PR #2584, opened 2021-05-28
Karr/fixes

jonrkarr:


bilalshaikh42:
It looks like this branch is based on the previous dev head.  I think the last commit (5482b90) is the only that is not already included in the dev branch. Could you create a new branch from the current dev and add that commit to it?

jonrkarr:
Will do. It's just 1 change to 1 line.

PR #2627, opened 2021-06-10
Use HSDS to read results

### LINKED ISSUE #2069: use the database for output, remove the connection to nfs
https://github.com/biosimulations/Biosimulations/blob/9913822809007492d191c47c21f6983d86e850e5/biosimulations/apps/dispatch-api/src/app/app.service.ts#L11-L16

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 9913822809007492d191c47c21f6983d86e850e5. It&#x27;s been assigned to @bilalshaikh42 because they committed the code.

### LINKED ISSUE #2369: move this conversion to the write side after parsing hdf files
https://github.com/biosimulations/Biosimulations/blob/3ab72df91be780e9b63a79c2be5233bbde13258e/biosimulations/apps/dispatch-api/src/results/results.controller.ts#L184-L189

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 3ab72df91be780e9b63a79c2be5233bbde13258e when #2358 was merged. cc @biosimulations.

### LINKED ISSUE #2440: Results cannot handle special characters in keys.
This has supposedly been resolved in later versions of mongo, but might need to set some options on the drivers.

[Some discussions on workarounds
](https://stackoverflow.com/questions/12397118/mongodb-dot-in-key-name)

### LINKED ISSUE #2442: Add attribute to results data model(s) to keep track of the type of each "report"
In order to render plots described in SED documents as discussed at the end of the HARMONY with Lucian Smith and Frank Bergmann, we need to save the data for plots (e.g. to HDF/CSV). After thinking about this, I think the simplest solution is to basically treat report and plot data the same -- save the plot data to CSV/HDF and use the same typescript/MongoDB data structures.

To help keep track of which data corresponds to `SedReport`s and which to `SedPlot2D` or `SedPlot3d`, it would helpful to record the type of the output associated with each instance of `ResultsModel` (`SedReport`, `SedPlot2D` or `SedPlot3d`). 

Example: add attribute with name `type` or `_type` to `ResultsModel` in `biosimulations/apps/dispatch-api/results/results.model.ts`. The value of this attribute could be `SedReport`, `SedPlot2D` or `SedPlot3D`.

(Ideally, we'd also rename attributes from `reportId` to `outputId`, reflecting their new more general meaning. But changing this seems like a lower priority.)

### LINKED ISSUE #2533: Fetch results from hsds instead of mongo
We can use the python client wrapped in flask to retrieve the results. however, this would be annoying to get working with the existing datamodel as they would need to be kept in sync manually and all the error handling etc would need to be impemented. 
Alternatively, we can use the  [OpenApi ](https://github.com/biosimulations/data-service-api/blob/openapi/openapi.yaml) spec to generate typescript client libraries.

bilalshaikh42:


bilalshaikh42:
> For symmetry, do you want to call this `sedmlId` and `sedmlName`, or change `sedmlId` and `sedmlName` of `SimulationRunReport` to be `id` and `name`?
> 
> Also should `reportId` of `SimulationRunReport` be `sedmlReportId` or is this now the same as `sedmlId`?

Changing everything to just id and name sounds better to me. No need to mention sedml here IMO. The reportId was actually renamed to outputId. If that is the same as sedmlId then that can be removed. I might structure the output id to be the full path, in which case I'll keep the sedmlId field as well

PR #2669, opened 2021-07-02
feat(dispatch,ontology): add terms for GINsim format

jonrkarr:
- Added EDAM terms for GINsim formats
- Added example simulation runs for GINsim, LibSBMLsim
- Debugged issues with OpenAPI spec for the COMBINE service


apps/combine-service/src/spec/spec.yml
```diff
@@ -417,6 +417,9 @@ paths:
                       master: true
                   files:
                   - <FILE>
+            encoding:
```
jonrkarr, line 420: @bilalshaikh42 in my experience, this needed to be added back. Please review.

bilalshaikh42: I was having some trouble with generating a client library with that field present. What is the reason it is needed? I am not familiar with the function of the "encoding" field.

bilalshaikh42: I see the issue. i did not realize that the encoding was needed for the input validation.

bilalshaikh42 (approved):

bilalshaikh42:
Merged

PR #2824, opened 2021-08-12
fix(dispatch-service): added handling for case when no environment va…

### LINKED ISSUE #2820: Simulation runs are broken
All simulation runs are broken. 
Seems to be caused by https://github.com/biosimulations/Biosimulations/commit/107221acfef0d611dba986f1da1e1782c1472d91 or one of the later commits

jonrkarr:
closes #2820


bilalshaikh42:
This seems have fixed it. Thanks!

PR #2863, opened 2021-08-26
feat(combine-service): isolating simulation execution into separate processes

### LINKED ISSUE #2851: Combine service container uses large amount of memory
The deployed container for the combine service uses up to 900Mi, causing the nodes to run out of memory.

jonrkarr:
closes #2851 

In my testing, this fixtures two issues
- Inability to run all simulators within one Python environment due to segmentation fault from interaction between NEURON and PySCeS
- Memory usage

Improved behavior
- I can now execute tests for all simulators from one parent Python environment (the individual simulators are run in subprocesses)
- Memory appears to be lower
- CPU usage is likely a little higher
- Latency is is likely slower

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
       - name: Install Flake8
         run: |
           python -m pip install flake8
-      
+
       - name: Lint
         run: nx affected:lint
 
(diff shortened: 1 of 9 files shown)


jonrkarr:
@bilalshaikh42 I think this is good to merge. The failures are because of insufficient memory for installing TypeScript for other apps.

What memory usage do you now see?

PR #2868, opened 2021-08-26
chore: disabled /run endpoints on production

jonrkarr:
This removes the `/run` endpoints from production deployment. Hopefully this will clear up the performance issue. If not, the problem must be with the Python environment inside the Docker image. In that case, there aren't many options except not to include the simulators in the Python environment, in which case we won't be able to run simulations.

@bilalshaikh42 will the following enable me to detect the deployment environment (i.e., will the value of the `ENV` environment variable be `dev` for deployment and `prod` for production)?

```python
os.environ.get('ENV', 'dev').lower() == 'prod'
```

### DIFF
--- a/apps/combine-service/src/app.py
+++ b/apps/combine-service/src/app.py
@@ -12,7 +12,7 @@ config = {
     **dotenv_values("secret/secret.env"),
     **dotenv_values("config/config.env"),
 }
-env = config.get("ENV", 'dev') || 'dev'
+env = config.get("ENV", 'dev') or 'dev'
 
 # disable ``/run`` endpoints from production
 if env.lower() == 'prod':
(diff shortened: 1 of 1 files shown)


bilalshaikh42:
The structure for the other applications is to use a config.env file to load in environment variables. That line will be able  detect the variables correctly, bit only if they are loaded in from the file. I do not remember if this functionality is already present in the combine service. If so, they should work and I can add the to the deployment. 

If not the package python-dotenv can be used to load the files. The files are mounted at config/config.env relative to the root.

bilalshaikh42:
I used the dotenv package to load the config variables in [this file](https://github.com/biosimulations/Biosimulations/blob/dev/apps/combine-service/src/s3.py).  The same code can be used to read the 'ENV' variable in the file that has the if statement.  
The [readme](https://github.com/theskumar/python-dotenv) for the package has a few additional ways of loading the variables, such as loading them into the environment so that the above line works. We could set it up so that we just load all the variables into the environment at once, but I just used the simple method for now. 
 I'll push the changes to the deployment file shortly

bilalshaikh42 (approved):

bilalshaikh42:
The changes to environment variables on the deployment are pushed

PR #2877, opened 2021-08-27
Various clean up of dispatch app and API

bilalshaikh42:


bilalshaikh42:
> ## replace below with above after sorting out the URI mapping. Move most of the mapping to the backend API
> https://github.com/biosimulations/Biosimulations/blob/700801798226e69b812595b562d46659bc5df3af/apps/dispatch/src/app/components/simulations/view/view.component.ts#L707-L710
> 
> ###### This comment was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in [7008017](https://github.com/biosimulations/Biosimulations/commit/700801798226e69b812595b562d46659bc5df3af) in #2877. cc @biosimulations.


@jonrkarr, could you take a look at this? I have used the dispatch API to get the metadata instead of the combine API, but I need to change the logic below the TODO. There are a few changes that need to be incorporated: 

1) The metadata validation will be happening when the project is published, so the fields that were guaranteed not to be null now could be. Because of this,  the metadata tab should be able to handle null values instead of showing the validation report.  If we do want to display the validation report to the user on this page to let them know if the simulation is publishable, we could separate it from the logic of parsing/displaying the metadata. 

2)  Ideally, I think we would remove the validation of the metadata from the combine-API entirely. The combine API can always return the correct metadata data model with 'null' values. Then the API can have the validation of which fields are needed for publication etc. This would greatly simplify the communication that needs to happen between the typescript and the python since the typescript code does not need to deal with error handling etc from the python API

3) The API now returns the simulation id in place of "." for the top-level URI and "id/xxx" instead of "./xxx" for elements in the archive. This means some of the mappings to the figures are off. I am having some trouble figuring out how the data sources are being mapped.

PR #2958, opened 2021-09-05
fixed: results URLs for charts

### LINKED ISSUE #2960: RunBioSimulations Vega and Plotly visualizations are broken
Requests for data are not properly URI encoded.

jonrkarr:
Closes #2960


bilalshaikh42:
TIL 
Thanks! Does the encode URIComponent need to be called at the input to the function as well?

bilalshaikh42:
Also the platform/account api are failing since I revoked the password they were using. Should be fine to merge in if those two are the only failures

jonrkarr:
I've also gotten `encodeURI` and `encodeURIComponent` mixed up. `encodeURI` sounds like what you want, but you actually want `encodeURIComponent`.

PR #2975, opened 2021-09-07
feat(dispatch-service): added passing software licenses from deployment secrets to Singularity run

### LINKED ISSUE #2973: Add environment variables for software licenses to singularity run in dispatch service
- [x] Organize environment variables in https://github.com/biosimulations/secrets. See biosimulations/secrets#1
- [x] Pull variables from secrets to sbatch service -- available as `process.env`
- [x] Add environment variables to singularity run at 
https://github.com/biosimulations/biosimulations/blob/dev/apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts#L44

jonrkarr:
Closes #2973 

I prefixed secrets to be passed to Singularity run with `SINGULARITY_RUN_ENV_VAR_`. These prefixes are stripped before being passed to Singularity run. Thus, new software licenses can be passed to Singularity run simply by adding environment variables with the above prefix to the `secrets` repository.

### DIFF
--- a/apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts
+++ b/apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts
@@ -41,24 +41,16 @@ export class SbatchService {
       apiDomain = 'https://run.api.biosimulations.dev/';
     }
 
-    const singularityRunEnvVars: EnvironmentVariable[] = Object.entries(process.env as {[key: string]: string})
-      .filter((keyVal: [string, string]): boolean => {
-        return keyVal[0].startsWith('SINGULARITY_RUN_ENV_VAR_');
-      })
-      .map((keyVal: [string, string]): EnvironmentVariable => {
-        return {
-          key: keyVal[0].substr('SINGULARITY_RUN_ENV_VAR_'.length),
-          value: keyVal[1],
-        };
-      });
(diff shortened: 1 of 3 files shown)


apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts
```diff
@@ -41,16 +41,24 @@ export class SbatchService {
       apiDomain = 'https://run.api.biosimulations.dev/';
     }
 
-    const envString = envVars.length
-      ? '--env ' +
-        envVars
+    const singularityRunEnvVars: EnvironmentVariable[] = Object.entries(process.env as {[key: string]: string})
```
bilalshaikh42: We should use the config service here rather than process.env

bilalshaikh42 (changes requested):

bilalshaikh42:
The environment variables should be loaded in by the config service, rather than through process.env. I'll make that change. 

> I prefixed secrets to be passed to Singularity run with SINGULARITY_RUN_ENV_VAR_. These prefixes are stripped before being passed to Singularity run.

Has this change been made in the secrets repo already? In the config service, I have been loading each secret individually. Is your intention that anything with that prefix would be loaded automatically?

bilalshaikh42 (approved):
@jonrkarr ready to merge on my end

bilalshaikh42:
merged

PR #3007, opened 2021-09-21
feat(platform): add visualization support to projects

bilalshaikh42:


bilalshaikh42:
@jonrkarr  I have gotten the vega visualizations working but am having a lot of rendering issues. The embedded vega element is very small unless the window is resized, triggering the event listener to re-render the figure you set up on the vega component.

I experimented with the various lifecycle hooks to get the render to trigger but have not had any luck.

jonrkarr:
This was tricky to workout.

My recollection is that I had to 

1. Set the `changeDetection` to `ChangeDetectionStrategy.OnPush`
    ```typescript
    @Component({
      templateUrl: './view.component.html',
      styleUrls: ['./view.component.scss'],
      //this seems to be required oddly
      changeDetection: ChangeDetectionStrategy.OnPush,
    })
    ```

2. Force Vega to update when visualization is opened, which I attached to an event handler for changing to the tab that the visualization is embedded in.

    ```typescript
     public selectedTabChange($event: MatTabChangeEvent): void {
        this.selectedTabIndex = $event.index;
        if ($event.index == this.iViewChartTab) {
          if (this.vegaVisualization) {
            this.vegaVisualization.render();
          }
        }
      }
    ```

3. Force change detection (`changeDetectorRef.detectChanges()`) when visualization is opened, which I attached to an event handler for changing to the tab that the visualization is embedded in.

    ```typescript
     public selectedTabChange($event: MatTabChangeEvent): void {
        this.selectedTabIndex = $event.index;
        if ($event.index == this.iViewChartTab) {
            this.changeDetectorRef.detectChanges();
        }
      }
    ```

jonrkarr:
Even the above isn't perfect. Vega still doesn't perfectly resize as intended all of the time.

bilalshaikh42:
I was wondering what that comment above the change detection was referring to. I have actually broken the visualization tab into a  separate component within the tab page, so I might need to find a way to pass the change tab event down to the component or look for some other event that can trigger the rendering.

jonrkarr:
I found a way to make it work in the dispatch app, but I don't feel I understand why it had to be that way. I tried various things until I found something that (mostly) works.

jonrkarr:
I believe you added the comment about change detection. 

You could try changing the change detection in the dispatch to see if that's really necessary.

jonrkarr:
The atypical change detection probably stems from Vega not being angular ui component. Plotly might be the only other ui component like this that we're using.

PR #3022, opened 2021-09-26
fix(dispatch): improved handling of undefined simulations in simulations browse view; closes #2999

### LINKED ISSUE #2999: Missing simulations in table view cause 404 page.
The user should only be directed to the 404 page if they try to look at a particular missing resource

The simulation run in the table should not trigger 404s

jonrkarr:
closes #2999

### DIFF
--- a/apps/dispatch/src/app/components/simulations/view/view.component.ts
+++ b/apps/dispatch/src/app/components/simulations/view/view.component.ts
@@ -6,7 +6,7 @@ import {
   ChangeDetectorRef,
   OnDestroy,
 } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import {
   FormBuilder,
   FormGroup,
@@ -228,6 +228,7 @@ export class ViewComponent implements OnInit, OnDestroy {
   constructor(
     private config: ConfigService,
     private route: ActivatedRoute,
+    private router: Router,
     private formBuilder: FormBuilder,
     private service: ViewService,
     private combineService: CombineService,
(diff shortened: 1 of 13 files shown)


apps/dispatch/src/app/components/simulations/view/view.service.ts
```diff
@@ -84,14 +87,14 @@ export class ViewService {
...
       id: simulation.id,
-      name: simulation.name,
-      simulator: simulation.simulator,
-      simulatorVersion: simulation.simulatorVersion,
-      cpus: simulation.cpus || 1,
-      memory: simulation.memory || 8,
-      maxTime: simulation.maxTime || 20,
-      envVars: simulation.envVars || [],
-      status: simulation.status,
+      name: simulation?.name as string,
```
bilalshaikh42, line 90: @jonrkarr. This effectively breaks type checking as type assertions override the compiler. I am not sure If this is the best approach to use here. Anything downstream of this method now needs to manually ensure that the data is actually a string and not null or undefined.

apps/dispatch/src/app/components/simulations/view/view.component.ts
```diff
@@ -299,9 +300,15 @@ export class ViewComponent implements OnInit, OnDestroy {
 
   private initSimulationRun(): void {
     this.Simulation$ = this.simulationService
-      .getSimulation(this.uuid)
+      .getSimulation(this.uuid)      
       .pipe(shareReplay(1));
 
+    this.Simulation$.subscribe((simulation: Simulation): void => {
```
bilalshaikh42, line 306: The additional subscription here is redundant with the formatted simulation below. We can just add this to the chain below to ensure that we don't pass undefined into the format simulation method, which would also avoid the type assertion problem

apps/dispatch/src/app/components/simulations/view/view.service.ts
jonrkarr: I realize this. That class is being used for two things. More work is needed to better separate this out.

PR #3030, opened 2021-09-27
chore(dispatch): improved handling of unknown simulation runs

### LINKED ISSUE #3031: Improve handling of unknown simulation runs
Address issues raises in #3022 which I lazily skipped due to the interwining of the simulation browse and view routes by the local storage of simulation runs in browsers.

> This effectively breaks type checking as type assertions override the compiler. I am not sure If this is the best approach to use here. Anything downstream of this method now needs to manually ensure that the data is actually a string and not null or undefined.

> The additional subscription here is redundant with the formatted simulation below. We can just add this to the chain below to ensure that we don't pass undefined into the format simulation method, which would also avoid the type assertion problem

jonrkarr:
closes #3031


jonrkarr:
@bilalshaikh42 I merged this so I could more easily start working from this. Open issue(s) for any additional work.

PR #3043, opened 2021-09-27
chore(platform): polish platform browse view

### LINKED ISSUE #3049: Lazy load images into platform browse simulations view

jonrkarr:
- Removed unused `projects` component
- Added lazy loading; closes #3049


bilalshaikh42 (commented):
The projects component was intended to be a landing page similar to the one we had created for the models earlier

jonrkarr:
Since the platform app now only has 1 section, this seems unnecessary. If we expand the app at some point, we can add something like this.

PR #3066, opened 2021-09-28
chore(dispatch): clean up metadata validation for publication

### LINKED ISSUE #3021: Link checking metadata validity for publication to storage of metadata in dispatch database
Relevant code: `this.metadataValid$` in `apps/dispatch/src/app/components/simulations/publish/publish.component.ts`

Currently, this is linked to the combine-service. This should be changed to the dispatch API.

- [ ] Dispatch data model must store errors about metadata (i.e. response from COMBINE API)
- [ ] Dispatch API must save these errors (alongside POST `/metadata/{simId}`)
- [ ] Dispatch API (GET `/metadata/{simId}`) must return these errors with code 400
- [ ] Publication component must check for the absence of such errors

jonrkarr:
closes #3021

### DIFF
--- a/apps/dispatch/src/app/services/dispatch/dispatch.service.ts
+++ b/apps/dispatch/src/app/services/dispatch/dispatch.service.ts
@@ -8,6 +8,7 @@ import {
   SimulationRun,
   UploadSimulationRun,
   UploadSimulationRunUrl,
+  UpdateSimulationRun,
   EnvironmentVariable,
 } from '@biosimulations/dispatch/api-models';
 import {
@@ -353,6 +354,15 @@ export class DispatchService {
         return of<undefined>(undefined);
       }),
     );
+  }
+
+  public updateSimulationRun(
+    uuid: string,
+    update: UpdateSimulationRun,
+  ) {
+    return this.http.patch(`${urls.dispatchApi}runs/${uuid}`, update, {
+      headers: { 'Content-Type': 'application/json' },
+    });
   }
 
   public constructor(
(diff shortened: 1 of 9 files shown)


bilalshaikh42:
Squash when merging to avoid adding merge commit

bilalshaikh42:
@jonrkarr I merged in the changes from dev/cleaned the committs.  Did you intend to wait for #3067  before merging this in?

apps/dispatch/src/app/services/dispatch/dispatch.service.ts
```diff
@@ -355,6 +356,15 @@ export class DispatchService {
     );
   }
 
+  public updateSimulationRun(
+    uuid: string,
+    update: UpdateSimulationRun,
+  ) {
+    return this.http.patch(`${urls.dispatchApi}runs/${uuid}`, update, {
```
bilalshaikh42, line 363: The dispatch app cannot use this method as it requires authentication. A token that allows this method to be called would allow any simulation to be updated and the token would need to be embedded in the front end. We can do something like this only  if we require user accounts to publish models

bilalshaikh42 (changes requested):

apps/dispatch/src/app/services/dispatch/dispatch.service.ts
jonrkarr: This was fine to merge. 

This is currently disabled. This is just a skeleton. This needs to be replaced with the final method to publish simulation runs.

PR #3074, opened 2021-09-28
feat(dispatch): require configuration of academic use for commercial solvers

### LINKED ISSUE #3059: Handle academic use of Gurobi
Issue discussed with @mpw6 @moraru 

- [x] Add academic use checkbox to simulation run form
- [x] Add academic use key to simulation run model
- [x] Only use Gurobi keys when academic use = true

jonrkarr:
closes #3059


bilalshaikh42:
Rather than having a boolean flag for academic use, lets create an enum such as tier or type that we can use to differentiate various different types. For now we can just have two options of acadmic/other but later can include others if it comes up

jonrkarr:
Feel free to use an enumeration instead, but it would be challenging to support multiple tiers and communicate what resources are available to which tiers. In practice, I think just two tiers is a good way to avoid overcomplicating this.

jonrkarr:
This is a big rabbit hole because the details of licenses can get quite nuanced
- Use
  - research
  - education
  - administrative
  - other
- Institutional setting
  - K-12
  - non-profit college/university
  - for profit college/university
  - non-profit research institute
  - government
  - corporation
- Country

If we were to go down that path, it would need to be an unordered enumeration of purposes. The dispatch-service would then have to have a map from purposes to eligible keys. The secrets would need to contain more information about which keys can be used for which purposes.

jonrkarr:
Switched to enumeration with two current values: `academic` and `other`.

In principle, the web form which collects the purpose and the dispatch-service could be expanded to other purposes. This would require explaining the licensing implications of this. However, I think its best to try to keep a small number of purposes to avoid users needed to read a lot of information to understand the implications of these choices.

bilalshaikh42 (approved):
Lgtm
Ill rebase the commits

bilalshaikh42 (approved):
Lgtm

Issue #2974, opened 2021-09-07
Copy model formats, simulation algorithms, and simulation task types from SED-ML to database entries for simulation runs


jonrkarr:
This would helpful information to users for filtering published projects. The COMBINE service will return this information. The service will return modelFormats as SED-ML URNs (e.g., `urn:sedml:language:sbml`). This can be mapped to EDAM ids (e.g., `format_2585`).

```typescript
interface simulationRun {
    ...
    modelFormats: EdamId[];
    simulationAlgorithms: KisaoId[];
   simulationTypes: ('SedUniformTimeCourseSimulation', 'SedSteadyStateSimulation', 'SedOneStepSimulation')[];
}
```

Rather than attempting to copy entire SED-ML documents into MongoDB, we could simply capture a small amount of key information as lists of strings (EDAM ids, KiSAO ids, and simulation types).

jonrkarr:
Here's an improvement which will better capture the requirements of the tasks (i.e., the combination of model formats, simulation types, and simulation algorithms that are used together within COMBINE archives).

### Data model
```typescript
interface Task {
    sedmlLocation: string;
    id: string;
    name: string;
    modelFormat: EdamId;
    simulationType: ('SedUniformTimeCourseSimulation' | 'SedSteadyStateSimulation' | 'SedOneStepSimulation');
    simulationAlgorithm: KisaoId;
}

interface simulationRun {
    ...
    tasks: Task[]; # list of unique requirements to run the tasks
}

const MODEL_LANGUAGE_TO_FORMAT: {[sedmlUrn: string}: EdamId} = {
  'urn:sedml:language:bngl': 'format_3972',   
  'urn:sedml:language:cellml': 'format_3240',   
  'urn:sedml:language:copasiml': 'format_9003',   
  'urn:sedml:language:ginml': 'format_9009',     
  'urn:sedml:language:hoc': 'format_9005',
  'urn:sedml:language:kappa': 'format_9006',
  'urn:sedml:language:lems': 'format_9004',
  'urn:sedml:language:mass': 'format_9011',
  'urn:sedml:language:morpheusml': 'format_9002',
  'urn:sedml:language:neuroml': 'format_3971',
  'urn:sedml:language:pharmml': 'format_9007',
  'urn:sedml:language:rba': 'format_9012',
  'urn:sedml:language:sbml': 'format_2585',
  'urn:sedml:language:smoldyn': 'format_9001',
[...]

PR #3089, opened 2021-10-02
fix(dispatch-service): fix processing of environment variables

### LINKED ISSUE #1: Create Header

### LINKED ISSUE #2: Create nav Bar

### LINKED ISSUE #3: Main layout

bilalshaikh42:


jonrkarr:
Is there something in particular you'd like me to look at? Otherwise, I'm inclined to say merge it and then let's test it end-to-end to make sure the changes work.

bilalshaikh42:
@jonrkarr still have some cleanup to do on this, but the following are done. 
1. process the manifest to get a list of files
2. save files to the database under the following endpoints.  File contains information about location, size, format, name
```
/files 
/files/simId
/files/simId/fileId
```
3.  Process the sedml specs to get the tasks, data generators, outputs, models and simulations
4. Save this to the database, available under the `specifications/simId`  endpoint

Currently I am saving all of the information in 3 to one endpoint called `/specifications`.  Is there some other organization that would be better?  For example #2974 suggests adding the simulations/algorithms info directly to the simulationRun object, which I agree makes more sense. Simmilarly we can separate out the visualizations/outputs since this can be used to quickly get all of the figures to display on a page for multiple runs without needing to download and parse the data.  

It would be easier to make these changes now (couple more hours) that later refactor front end and endpints

jonrkarr:
Saving all of the SED-ML specs together is good. This parallels the COMBINE service. In that case, we will simply be able to change the projects service in the platform app from the COMBINE api to the new endpoint. Lets start with that.

bilalshaikh42:
> Saving all of the SED-ML specs together is good. This parallels the COMBINE service. In that case, we will simply be able to change the projects service in the platform app from the COMBINE api to the new endpoint. Lets start with that.

Sounds good. I just realized since the specs are all one document, I can just add some endpoints that return a portion of it, so this would be an easier change to make later rather than actually storing them as separate documents.  

The main question then is what to expose through the API. For the files for example, I am hiding all of the complexity of things like `_type: CombineArchiveContentLocation` etc and just outputting the location as a string. What are the key fields of the sedml specs that are needed?

I'll clean this up and merge, and then work on getting the simulation info (algorithms/formats etc) in a new PR to include in the simulation run object.

jonrkarr:
There's two things we'd ideally do next to reduce the data we're pulling from the API to populate

Reasons for processing out information
- List of visualizations in the view route
- Filter simulations in the browse route on additional attributes not currently available in the SimulationRun or metadata

Data to process out
- Process out the visualizations
  - Create a list of visualization objects
  - Create an endpoint to return the ids/names of the visualizations
  - Create an endpoint to return the details of an individual visualization -- this could parallel these data models: https://github.com/biosimulations/biosimulations/blob/polish-published-simulation-view/libs/datamodel/view-simulation/src/lib/view-simulation.ts#L6
- Process out some additional simulation run metadata
  - From SED-ML specs
    - Set of model formats (e.g., `urn:sedml:langauge:sbml`)
    - Set of simulation types (e.g., uniform time course)
    - Set of algorithms (KISAO ids)
  - From logs
    - Duration
    - Set of executed algorithms

Thoughts for organizing this
- Visualizations: Separate, new endpoints
- Simulation run metadata: no strong thoughts here. This could be returned via a separate, new endpoint for simulation run metadata. This would distinguished from SimulationRun which contains information about the requested run. Also, this would be read only information (unlike SimulationRun) derived from the submitted COMBINE archive.

jonrkarr:
I agree these new endpoints could join and slice information from existing documents. I agree we don't need additional documents. Using the existing documents would make it easier to pull out additional information going forward without needing to migrate the documents. If we pull information from existing documents, the core storage should be pretty stable because it just maps to the COMBINE archive, OMEX manifest, SED-ML, RDF, log, and results formats. Most of these are pretty stable and we haven't been changing our log or results format either.

The endpoint for returning visualizations could do this
- Vega: read the file from S3 and return it
- SED-ML read a portion of the document for the whole specs

The same could be done for the other simulation run metadata (dynamically read this from existing documents) rather than storing this information twice.

Issue #3083, opened 2021-09-29
Add typing for simulations in platform view service


jonrkarr:
This is due to the scoping of the data models for the dispatch-api which prevents importing into the platform app.

PR #3153, opened 2021-10-06
refactor(combine-service): separated types for manifest and sed-ml-specs endpoints

### LINKED ISSUE #3091: Distinguish response types for each combine service endpoint
I might be reading this incorrectly, but it seems that the manifest endpoint and sedml  specs endpoint are both returning the same content 

https://combine.api.biosimulations.org/ui/#/COMBINE/src.handlers.combine.get_sedml_specs_for_combine_archive.handler

jonrkarr:
Closes #3091

Changed the type for two endpoints:
- `combine/manifest`
  - `CombineArchive` --> `CombineArchiveManifest`
  - `CombineArchiveContent` --> `CombineArchiveManifestContent`
  - `CombineArchiveLocation` --> `CombineArchiveManifestLocation`
- `combine/sedml-specs`
  - `CombineArchive` --> `CombineArchiveSedDocSpecs`
  - `CombineArchiveContent` --> `CombineArchiveSedDocSpecsContent`
  - `CombineArchiveLocation` --> `CombineArchiveSedDocSpecsLocation`

The new types are copies of the existing types, except the types of `CombineArchiveManifestLocation.value` and `CombineArchiveSedDocSpecsLocation.value`. Their types are now different, and each is a single type rather. Previously, their types were `anyOf` their two types.

I think the revised COMBINE service can be used the with TypeScript apps as is. I don't think the TypeScript apps ever check the `_type` attributes for these classes.

@bilalshaikh42 let me know how you're thinking of modifying the TypeScript apps/libs in response to this.

Note, `SedDocument` is still used by the three endpoints below. I think this makes sense. The first two allow writing and reading SED-ML documents. The third essentially generates SED-ML documents from model files. The third still only uses a subset of SED-ML, but its moving toward using all of SED-ML.
- `/combine/create`
- `/combine/sedml-specs`
- `/sed-ml/get-parameters-variables-for-simulation`


bilalshaikh42:
Thanks! I'll regenerate the typescript code and hopefully everything should just work

jonrkarr:
@bilalshaikh42 I think we should add an issue to update some of the UI code to use these newer, stricter types.

bilalshaikh42:
These types are only going to be used by the API correct? The API will then return types that are used on the front end?
The files for example look quite different from the API that from the  combine maifest

jonrkarr:
Yes, but I think the typing in the new `libs/projects-service` still need some clarification.

bilalshaikh42:
> Yes, but I think the typing in the new `libs/projects-service` still need some clarification.

Yup should be covered by #3083 and #3101

PR #3172, opened 2021-10-07
feat(dispatch): enhanced validation

jonrkarr:
- [x] Add API endpoints for additional and expanded validation
- [x] Update `cOMBINE` service and `combine` service
- [x] Correct metadata checking for publication
- [x] Add options to project validation form
- [x] Add explanation to project validation form
- [x] Add model validation form
- [x] Add simulation experiment (SED-ML) validation form
- [x] Add metadata (OMEX Metadata) validation form
- [x] Debug validation with 3 model languages (likely in BioSimulators-utils)
  - [x] BNGL -- interesting issue; the BNGL library doesn't raise errors on parsing totally different formats that it has no ability to parse. Instead, they're read as something like an empty model.
  - [x] NeuroML -- similar issue to BNGL
  - [x] XPP -- my error detection was overly strict. I looked for "ERROR" (upper case). Apparently XPP doesn't always use upper case.


jonrkarr:
@bilalshaikh42 this is good to merge. I'll take care of the remaining issues separately. Those are likely in BioSimulators-utils are won't affect the TypeScript libraries/apps.

jonrkarr:
@bilalshaikh42 all set to merge. Model validation improved.

jonrkarr:
I merged because updating your COMBINE client was straightfoward. This is very helpful for managing usage of the COMBINE service.

PR #3182, opened 2021-10-08
refactor(platform): added tool to publish example

jonrkarr:
- Indicated in which examples should be published and what ids to use in `tools/example-projects.json`
  - This avoids polluting our published projects with many versions of the same project with different simulation tools
- Added command-line program to update these published projects
- Presently requires a machine-to-machine api client and secret.

### DIFF
--- a/tools/example-projects.json
+++ b/tools/example-projects.json
@@ -60,7 +60,7 @@
     "filename": "rba/Escherichia-coli-K12-WT.omex",
     "disabled": false,
     "publishToBioSimulations": true,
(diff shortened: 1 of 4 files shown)


bilalshaikh42:
Thanks! 
We can keep it as a m2m or make it an interactive login since we will be running it manually

jonrkarr:
I started with machine-to-machine because its easier to setup. I could append this to the GitHub action I created to check the examples and automatically create/update their publication.

PR #3186, opened 2021-10-08
feat(platform): added validation for project ids

### LINKED ISSUE #3183: Add validation for projects
- [x] Id matches ^[a-zA-Z0-9_-]{3,}$
- [ ] simulationRun is the id of a valid run
- [ ] The status of that run was successful
- [ ] That run has at least one root level metadata document (uri with no slashes in it)

The UI does all such checking. But it should also be enforced at the database level so that the API cannot bypass these checks.

jonrkarr:
closes #3183


bilalshaikh42:
It might be better to add this validation to the controller level instead

jonrkarr:
I leave it to you to move the validation

bilalshaikh42:
> I leave it to you to move the validation

I kept both. The database validation is helpful for preventing mistakes made by the API code and the controller is user facing.

PR #3237, opened 2021-10-16
feat(dispatch,simulators,platform): prominently display down status on apps

### LINKED ISSUE #3210: Add status badges/links to app UIs
- [x] Pull status from new `/health` endpoints
- [x] Bottom of all pages
- [x] Main menus of apps

jonrkarr:
closes #3210

When a status of a relevant service to an app is `down`, displays a red bar at the bottom of each page of the app. The bar is displayed similar to the privacy policy notice.

Todo
- [ ] Link to new `/health` endpoint after #3308 is finished


bilalshaikh42:
I don't want to end up tying the structure of the status checks to the app. This is bound to break and cause lots of issues. The status checks were set quite quickly so more thought would be needed on what portions/names to test. 
One alternative is to just go through all the statuses and show the error if anything is down.

apps/simulators/src/app/app.component.ts
```diff
@@ -1,19 +1,31 @@
...
+  status$!: Observable<boolean>;
+
   constructor(
     public config: ConfigService,
     private scrollService: ScrollService,
+    private systemStatusService: SystemStatusService,
   ) {}
 
+  ngOnInit(): void {
+    this.status$ = this.systemStatusService.getAppStatus([
```
bilalshaikh42: Using as here is bound to cause issues If any network requests fail. In general using as should be avoided as all checks for null/undefined are lost. This leads to errors such as the one caused by the simulators response which would otherwise be caught by the compiler

libs/datamodel/common/src/lib/common/system-status.ts
```diff
@@ -0,0 +1,10 @@
+export interface SystemStatus {
```
bilalshaikh42: I really don't think this should be included in the data model. This will be adding another external dependency that now needs to be kept in sync. If we really want to have it, it should at least be completely encapsulated by the service and not exported for use by other apps

bilalshaikh42 (changes requested):

bilalshaikh42:
Closing in favor of  a solution that utilizes #3308 
Health endpoint will return either 200 or 503, which is sufficient to point users to status.biosimulations.org for details

jonrkarr:
Most of this PR doesn't directly relate to where status is checked. That part can be replaced. 

While I did define an interface, I chose not to make the apps use this. Somewhere the apps need to declare the services that are relevant to them.

libs/datamodel/common/src/lib/common/system-status.ts
jonrkarr: Removed this logic here in favor of moving it to the health module.

libs/api/angular-client/src/lib/health/health.service.ts
```diff
@@ -0,0 +1,30 @@
...
+// import { HttpErrorResponse } from '@angular/common/http';
+
+@Injectable()
+export class HealthService {
+  private endpoints = new Endpoints();
+  
+  constructor(private http: HttpClient) {}
+
+  isHealthy(): Observable<boolean> {
+    return of(true);
```
jonrkarr: Get health here

nx.json
```diff
@@ -86,6 +86,7 @@
     "api-angular-client": {
       "tags": [
         "scope:simulations",
+        "scope:simulators",
```
jonrkarr: The scope of the angular-api-client had to be broadened to allow the simulators app to also get health information from the single health endpoint that will be added to the BioSimulations API through the angular-api-client.

bilalshaikh42:
the simulators endpoint will need its own health endpoint anyway, so we can
use that. I would really like to avoid broadening the scope as then the
simulators app will have dependencies on the main API which we definitely
do not want

jonrkarr:
I agree there should be different health information for different apps. This is what I was starting to set up. I changed this because you discussed 1 health endpoint.

nx.json
jonrkarr: I separated health checking for the two APIs

bilalshaikh42:
We should now have the following deployed: 

API: /health/status 
SimAPI: /health 

the /health on the API is not complete, meaning it does not check for other dependencies that may have issues. this is to prevent kubernetes from killing the API just because the database is down and triggering a failure cascade. The front end should use /health/status to get a more accurate status from the user's perspective

apps/platform/src/app/app.component.spec.ts
```diff
@@ -20,14 +21,14 @@ describe('AppComponent', () => {
...
+      providers: [AuthService, AuthEnvironment, ConfigService, ScrollService, HealthService],
       declarations: [AppComponent],
     }).compileComponents();
   }));
 
   it('should create the app', () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
-    expect(app).toBeTruthy();
+    // const fixture = TestBed.createComponent(AppComponent);
```
bilalshaikh42, line 39: We can just mock the health service to prevent any issues

apps/simulators/src/app/app.component.spec.ts
```diff
@@ -20,14 +21,14 @@ describe('AppComponent', () => {
...
-      providers: [ConfigService, ScrollService],
+      providers: [ConfigService, ScrollService, HealthService],
     }).compileComponents();
   }));
 
   it('should create the app', () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
-    expect(app).toBeTruthy();
+    //const fixture = TestBed.createComponent(AppComponent);
```
bilalshaikh42, line 31: Same here

libs/config/common/src/lib/endpoints.ts
```diff
@@ -11,11 +11,13 @@ export class Endpoints {
   private api: string;
```
bilalshaikh42: Changes in this file are already done in #3308 so can be reverted

bilalshaikh42 (changes requested):

apps/platform/src/app/app.component.spec.ts
jonrkarr: This is because of the updater.

apps/simulators/src/app/app.component.spec.ts
jonrkarr: Same as above

apps/platform/src/app/app.component.spec.ts
bilalshaikh42: The update service for the PWA?

libs/config/common/src/lib/endpoints.ts
jonrkarr: I removed the endpoint method I created.

I needed to correct the simulators endpoint method.

jonrkarr:
Ready to merge. 

The status banner displays once the response from the health endpoint fails. If there's a delay, the banner will pop up after a short delay. When the API is working, users won't see the banner.

The banner appears at the bottom of each page and can be dismissed. The banner includes a link to https://status.biosimulations.org.

jonrkarr:
Lint failure fixed in another branch.

PR #3262, opened 2021-10-17
fix(api,hsds): update hsds client library

bilalshaikh42:
fixes build errors, simplifies library


jonrkarr:
Thanks! I was just going to try to fix this.

PR #3264, opened 2021-10-17
Update README.md

bilalshaikh42:
Experimenting with making readme more friendly


jonrkarr (approved):
Fine with me. Seems like reasonable choices. I wish GitHub had less cartoonish options.

PR #3318, opened 2021-10-19
feat(dispatch): recorded simulator versions and digests for simulation runs

### LINKED ISSUE #3195: SimulationRuns should store simulator version used rather than `latest`
If a run is requested to use the "latest" version, the actual version used should get stored (rather than the request for the latest version). This is important for reproducibility.

In addition, the digest (hash) of the simulator should always be stored for reproducibility. This is important because we're allowing developers to overwrite image tags. As a result, we should save the image digest, which is unique to a specific version of a tag for a Docker image.

Ideally, we would get the digests from the SBatch service. However, Singularity doesn't seem to make it easy to get digests. Singularity keeps this information in its image cache, but doesn't store docker image tags and digests together. See https://sylabs.io/guides/3.0/user-guide/singularity_and_docker.html?highlight=digest#oci-image-support

Instead, we can collect the digests from the simulators database. Assuming there aren't problems with image refreshing and simulations aren't initiated during image refreshing, the two solutions would return the same digests.

- [x] Verify requested simulator/version
- [x] Save version of simulation to simulation run
- [x] Save digest of image for simulator/version to simulation run

jonrkarr:
closes #3195

- Uses simulators database to validate requested simulator/version
- Raises BadRequestException if an image isn't registered for the requested simulator/version
- Saves digest (hash) for simulator/version to `SimulationRun`
- Added common data models for simulators and simulation runs
- Added simulatorDigest to SimulationRunUpdate. I added this when thinking about using the SBatch service to get image digests. This could be removed. I ended up not using this and instead getting digests from the simulators database.

@bilalshaikh42 the key thing to review is the integration of the simulator API (L307 of `apps/api/src/simulation-run/simulation-run.service.ts`).

### DIFF
--- a/apps/api/src/simulation-run/simulation-run.service.ts
+++ b/apps/api/src/simulation-run/simulation-run.service.ts
@@ -12,6 +12,7 @@ import {
   Logger,
   NotFoundException,
   PayloadTooLargeException,
+  BadRequestException,
 } from '@nestjs/common';
 import { InjectModel } from '@nestjs/mongoose';
 
@@ -28,7 +29,7 @@ import {
   UploadSimulationRun,
   UploadSimulationRunUrl,
 } from '@biosimulations/datamodel/api';
-import { SimulationRunStatus } from '@biosimulations/datamodel/common';
+import { SimulationRunStatus, ISimulator } from '@biosimulations/datamodel/common';
 import { SimulationStorageService } from '@biosimulations/shared/storage';
 import {
   DispatchFailedPayload,
@@ -40,6 +41,8 @@ import { BiosimulationsException } from '@biosimulations/shared/exceptions';
 import { Readable } from 'stream';
 import { firstValueFrom } from 'rxjs';
 import { DeleteResult } from 'mongodb';
+import { Endpoints } from '@biosimulations/config/common';
+
 // 1gb in bytes to be used as file size limits
 const ONE_GIGABYTE = 1000000000;
 const toApi = <T extends SimulationRunModelType>(
@@ -52,6 +55,7 @@ const toApi = <T extends SimulationRunModelType>(
 
 @Injectable()
 export class SimulationRunService {
+  private endpoints = new Endpoints();
   private logger = new Logger(SimulationRunService.name);
 
   public constructor(
@@ -236,6 +240,7 @@ export class SimulationRunService {
       );
     }
   }
+
   public async createRunWithURL(
     body: UploadSimulationRunUrl,
   ): Promise<SimulationRunModelReturnType> {
@@ -298,6 +303,15 @@ export class SimulationRunService {
     id: string,
   ): Promise<SimulationRunModelReturnType> {
     const newSimulationRun = new this.simulationRunModel(run);
+
+    const url = this.endpoints.getSimulatorsEndpoint(run.simulator, run.simulatorVersion);
+    const simulatorResponse = await firstValueFrom(this.http.get<ISimulator | null>(url));
+    if (simulatorResponse && simulatorResponse.data && simulatorResponse.data.image) {
+      newSimulationRun.simulatorVersion = simulatorResponse.data.version;
+      newSimulationRun.simulatorDigest = simulatorResponse.data.image.digest;
+    } else {
+      throw new BadRequestException(`No image for ${run.simulator}:{run.simulatorDigest} is registered with BioSimulators.`);
+    }
 
     const session = await this.simulationRunModel.startSession();
     // If any of the code within the transaction fails, then mongo will abort and revert any changes
--- a/apps/api/src/simulation-run/simulation-run.service.ts
+++ b/apps/api/src/simulation-run/simulation-run.service.ts
@@ -44,6 +44,7 @@ import { catchError } from 'rxjs/operators';
 import { DeleteResult } from 'mongodb';
 import { Endpoints } from '@biosimulations/config/common';
 import { HttpErrorResponse } from '@angular/common/http';
+import { ConfigService } from '@nestjs/config';
 
 // 1gb in bytes to be used as file size limits
 const ONE_GIGABYTE = 1000000000;
@@ -57,7 +58,7 @@ const toApi = <T extends SimulationRunModelType>(
 
 @Injectable()
 export class SimulationRunService {
-  private endpoints = new Endpoints();
+  private endpoints: Endpoints;
   private logger = new Logger(SimulationRunService.name);
 
   public constructor(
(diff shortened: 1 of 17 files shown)


apps/api/src/simulation-run/simulation-run.service.ts
```diff
@@ -299,6 +304,15 @@ export class SimulationRunService {
   ): Promise<SimulationRunModelReturnType> {
     const newSimulationRun = new this.simulationRunModel(run);
 
+    const url = this.endpoints.getSimulatorsEndpoint(run.simulator, run.simulatorVersion);
```
jonrkarr: Here's where the requested simulator/version is checked and the digest is obtained.

apps/api/src/simulation-run/simulation-run.service.ts
```diff
@@ -52,6 +55,7 @@ const toApi = <T extends SimulationRunModelType>(
 
 @Injectable()
 export class SimulationRunService {
+  private endpoints = new Endpoints();
```
bilalshaikh42: Need to pass in the env here from the config service

libs/datamodel/api/src/lib/core/simulationRun.ts
```diff
@@ -267,6 +277,14 @@ export class PatchSimulationRun {
   })
   public?: boolean;
 
+  @ApiPropertyOptional({
+    description: 'Digest of the simulation tool for the simulation run',
```
bilalshaikh42: Do we want the digest to be changeable?

apps/api/src/simulation-run/simulation-run.service.ts
```diff
@@ -299,6 +304,15 @@ export class SimulationRunService {
   ): Promise<SimulationRunModelReturnType> {
     const newSimulationRun = new this.simulationRunModel(run);
 
+    const url = this.endpoints.getSimulatorsEndpoint(run.simulator, run.simulatorVersion);
+    const simulatorResponse = await firstValueFrom(this.http.get<ISimulator | null>(url));
+    if (simulatorResponse && simulatorResponse.data && simulatorResponse.data.image) {
+      newSimulationRun.simulatorVersion = simulatorResponse.data.version;
```
bilalshaikh42: Why are we overriding the version here? Is it possible this is different from what the user requested?

apps/api/src/simulation-run/simulation-run.service.ts
bilalshaikh42: The check and info looks good.

bilalshaikh42 (changes requested):
The check and getting digest looks good.

libs/datamodel/api/src/lib/core/simulationRun.ts
bilalshaikh42: I see your comment about this now. I agree we should remove this as we don't want to have this field change

apps/api/src/simulation-run/simulation-run.service.ts
jonrkarr: This translates "latest" into a specific version. If the user requests and unregistered version, they would get an error.

libs/datamodel/api/src/lib/core/simulationRun.ts
jonrkarr: I'll remove this. This was because I was thinking about getting the digest from the Singularity image itself. Mainly to cover the possible case that the simulators database has diverged from the HPC environment because the refreshing of an image failed. Getting the digest from Singularity doesn't appear to be straightforward.

libs/api/nest-client/src/lib/simulation-run.service.ts
```diff
@@ -73,7 +73,7 @@ export class SimulationRunService {
       map((token) => {
         const httpRes = this.http
           .patch<SimulationRun>(
-            `${this.endpoint}run/${id}`,
```
jonrkarr, line 76: @bilalshaikh42 this seemed incorrect. Shouldn't this be `runs` plural?

apps/api/src/simulation-run/simulation-run.service.ts
jonrkarr: Done.

@bilalshaikh42 most uses of `Endpoints` don't pass an environment. Is the argument only needed for use in servers?

libs/datamodel/api/src/lib/core/simulationRun.ts
jonrkarr: Removed.

apps/api/src/simulation-run/simulation-run.service.ts
bilalshaikh42: Yup. On the front ends, the endpoints class will use the environment as defined in the environment file, since it is only changeable  at compile time. 

On the server, loading the environment dynamically is possible, so we can set an environment variable that can be read from the config service.

apps/api/src/simulation-run/simulation-run.service.ts
```diff
@@ -317,6 +339,52 @@ export class SimulationRunService {
     return toApi(newSimulationRun);
   }
 
+  private getSimulator(simulator: string, simulatorVersion: string): Promise<ISimulator | null | false> {
```
jonrkarr: @bilalshaikh42 this needs to be tested. I can't test because I can't run the API.

apps/api/src/simulation-run/simulation-run.service.ts
bilalshaikh42: That makes sense

libs/api/nest-client/src/lib/simulation-run.service.ts
bilalshaikh42: Either works to maintain compatibility with the older version. But yes, runs is preferred

apps/api/src/simulation-run/simulation-run.service.ts
bilalshaikh42: will do

bilalshaikh42: Does not seem to be working. I am working on debugging it. What is the issue with running the API? The docker image/compose file in the .dev container should be an easy way to set up the environment. If you open the directory with vscode it should prompt if you want to open up the repo in the dev container

apps/api/src/simulation-run/simulation-run.service.ts
jonrkarr: Basically, this is to make the provenance of simulation results more concrete.

bilalshaikh42:
@check-spelling-bot apply changes.

bilalshaikh42 (approved):

PR #3325, opened 2021-10-20
refactor(api): commented out unused buggy endpoint

### LINKED ISSUE #3240: Simplify or debug `/runs` endpoint and clarify documentation
- Simplify endpoint to simply return a list of ids of simulation runs
- OR debug the endpoint
  - [ ] Endpoint fails with multiple query fields.
  - [ ] The method seems to only work for string-valued fields.
- Clarify documentation
  - [ ] Clarify documentation of query argument
  - [ ] Clarify documentation of response type

jonrkarr:
closes #3240

Also
- Removed redundant endpoint methods
- Clarified names of arguments of endpoint methods; made consistent with terminology in Swagger documentation

### DIFF
--- a/apps/api/src/simulation-run/simulation-run.controller.ts
+++ b/apps/api/src/simulation-run/simulation-run.controller.ts
@@ -9,7 +9,7 @@ import { DispatchJob } from '@biosimulations/messages/messages';
 import { OptionalAuth, permissions } from '@biosimulations/auth/nest';
 import {
   ErrorResponseDocument,
-  FieldsQueryParameters,
+  // FieldsQueryParameters,
 } from '@biosimulations/datamodel/api';
 import {
   BadRequestException,
@@ -27,7 +27,7 @@ import {
   UploadedFile,
   UseInterceptors,
   UnsupportedMediaTypeException,
-  Query,
+  // Query,
 } from '@nestjs/common';
 import { FileInterceptor } from '@nestjs/platform-express';
 import {
@@ -37,7 +37,7 @@ import {
   ApiOkResponse,
   ApiNoContentResponse,
   ApiOperation,
-  ApiQuery,
+  // ApiQuery,
   ApiPayloadTooLargeResponse,
   ApiTags,
   ApiNotFoundResponse,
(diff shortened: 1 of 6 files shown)


apps/api/src/simulation-run/simulation-run.controller.ts
```diff
@@ -91,12 +91,14 @@ export class SimulationRunController {
...
     return (<multipartSimulationRunBody>body).simulationRun != undefined;
   }
+
   private isUrlBody(
     body: multipartSimulationRunBody | UploadSimulationRunUrl,
   ): body is UploadSimulationRunUrl {
     return (<UploadSimulationRunUrl>body).url != undefined;
   }
 
+  /*
```
bilalshaikh42: This method is used when developing and testing.  Since it requires authentication, it should not be confusing for users/reviewers as they will not have permission to call it. I would prefer to leave it in here

apps/api/src/simulation-run/simulation-run.service.ts
```diff
@@ -22,7 +22,7 @@ import {
   SimulationRunModel,
   SimulationRunModelReturnType,
   SimulationRunModelType,
-  SimulationRunField,
+  // SimulationRunField,
```
bilalshaikh42: Same here with leaving in in for the above method to work

bilalshaikh42 (changes requested):

apps/api/src/simulation-run/simulation-run.controller.ts
jonrkarr: I replaced this with a simpler method that returns the list of simulation runs with all attributes.

Alternatively, the method could be debugged to properly return selected fields indicated via query arguments.

apps/api/src/simulation-run/simulation-run.service.ts
jonrkarr: Updated accordingly

apps/api/src/simulation-run/simulation-run.controller.ts
bilalshaikh42: I would prefer to debug it as some point and eventually add consistent options to select specific fields to all the endpoints

jonrkarr: Let's merge this and create a separate issue.

Issue #3164, opened 2021-10-07
Change SimulationRun object to take in a BiosimulatorsId instead of a simulator/version pair


bilalshaikh42:
@jonrkarr  this would allow the database the store the biosimulators id directly so that there is no need to do mapping. This way those who are using the API to pull the simulation run would actually receive an id to an existing simulator object.

 There would not be much change required to the UI,  just mapping the current fields to the new term
The only tricky part would be mapping these ids to the correct images on the HPC, but this should be possible. I just went with the method akhil had when he set it up, but it should be easy enough to work out.

Thoughts?

jonrkarr:
What is a `BiosimulatorsId`? Object with two fields `id` and `version`?

bilalshaikh42:
Sorry, that was clear. basically just simulator/version as one string. I
know it can be easily reconstructed, but it just seems to make more sense
in the backend to have one id for each simulator

bilalshaikh42:
But now I am unclear on one thing. What was the mapping that you mentioned
needing to do because of the way that the simulator name and version were
stored? That was the initial issue that made me think about this, but now I
don't remember if there was any sort of problem

jonrkarr:
This is the information I'm trying to display in the dispatch and platform apps (more specifically in a component in the dispatch app and a component in `shared/project-ui`)
- Name of simulator
- Version of simulator

Presently, I'm getting the simulator `id` and `version` from the `SimulationRun` and then using the simulators API to get the `name` of the simulator from its `id`.

This could be factored differently.

jonrkarr:
A combined id field would be fine internally. I'd like to keep separate fields with the individual id and version attributes as well because we don't necessarily want to display the two together in the UI, and it would be annoying to have to parse that out. Also, I think contributors should submit the two pieces of information separately because they should be focused on registering ids for simulation tools.

The revised data model could look like this:
- `id`: read/write; no change
- `version`: read/write; no change
- `digest`: read/write; no change
- `idVersion`: read only, new attribute (e.g., `${id}:${version}`).

For simulation runs, a combined id/version attribute could be added and all of the simulator properties could be grouped together
- `simulator`:
  - `idVersion`
  - `id`
  - `version`
  - `digest`

bilalshaikh42:
From the users point of view, the simulators and versions make more sense as being separate.

PR #3335, opened 2021-10-20
feat(simulators): added combined id/version property

jonrkarr:
First steps toward #3164 

`idVersion` property is set just before saving/updating to database because I don't believe MongoDB has a great way to handle this. Pre-save hooks don't cover updating. Functions for defaults only apply when the value isn't directly set. Virtual properties can't be searched.

The simulators database could be migrated with the Python script below. I already ran this for the `dev` database. The first step saves a copy locally in case of any of problems.
```python
import json
import requests
import subprocess

response = requests.get('https://api.biosimulators.org/simulators?includeTests=true')
response.raise_for_status()
simulators = response.json()
with open('simulators.json', 'w') as file:
    json.dump(simulators, file)

response2 = requests.post('https://auth.biosimulations.org/oauth/token', json={
    'client_id': '...',
    'client_secret': '...',
    'audience': 'api.biosimulators.org',
    "grant_type": "client_credentials",
})
response2.raise_for_status()
response_data = response2.json()
auth = {'Authorization': response_data['token_type'] + ' ' + response_data['access_token']}

response2 = requests.delete(
    'https://api.biosimulators.org/simulators', json=simulator, headers=auth)
response2.raise_for_status()

for simulator in simulators:
    print(simulator['id'])

    response2 = requests.post(
        'https://api.biosimulators.org/simulators', json=simulator, headers=auth)
    response2.raise_for_status()
```

### DIFF
--- a/libs/config/common/src/lib/urls.ts
+++ b/libs/config/common/src/lib/urls.ts
@@ -36,7 +36,7 @@ const envUrls: { [key in envs]: urlMap } = {
     accountApi: 'https://account.biosimulations.dev/',
     dispatchApi: 'https://run.api.biosimulations.dev/',
     combineApi: 'https://combine.api.biosimulations.dev/',
-    simulatorsApi: 'https://api.biosimulators.dev/',
+    simulatorsApi: 'http://localhost:3333/',
     platform: 'https://biosimulations.dev',
     account: 'https://login.biosimulations.dev',
     dispatch: 'https://run.biosimulations.dev',
--- a/libs/config/common/src/lib/urls.ts
+++ b/libs/config/common/src/lib/urls.ts
@@ -36,7 +36,7 @@ const envUrls: { [key in envs]: urlMap } = {
     accountApi: 'https://account.biosimulations.dev/',
     dispatchApi: 'https://run.api.biosimulations.dev/',
     combineApi: 'https://combine.api.biosimulations.dev/',
-    simulatorsApi: 'http://localhost:3333/',
+    simulatorsApi: 'https://api.biosimulators.dev/',
     platform: 'https://biosimulations.dev',
     account: 'https://login.biosimulations.dev',
     dispatch: 'https://run.biosimulations.dev',
--- a/libs/simulators/database-models/src/lib/simulator.ts
+++ b/libs/simulators/database-models/src/lib/simulator.ts
@@ -36,10 +36,20 @@ export class Simulator extends Document implements ISimulator {
 
   @Prop({
     type: String,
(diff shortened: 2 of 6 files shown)


bilalshaikh42:
I am bit unclear as to what the purpose of this is.

libs/simulators/database-models/src/lib/simulator.ts
```diff
@@ -34,12 +34,22 @@ export class Simulator extends Document implements ISimulator {
...
+  })
+  idVersion!: string;
+
   @Prop({
     type: String,
     lowercase: true,
     trim: true,
     required: true,
     default: undefined,
+    index: true,
```
bilalshaikh42, line 52: We don't want to have this index here because we are setting a compound index below

libs/simulators/database-models/src/lib/simulator.ts
```diff
@@ -152,7 +163,7 @@ export class Simulator extends Document implements ISimulator {
 export const SimulatorSchema = SchemaFactory.createForClass(Simulator);
 
 // Can not be set in the decorator for compund schemas.
-SimulatorSchema.index({ id: 1, version: 1 }, { unique: true });
+SimulatorSchema.index({ id: 1, version: 1 }, { unique: false });
```
bilalshaikh42, line 166: Why are we setting this to false? Each entry should have a unique id/version combination correct?

bilalshaikh42 (changes requested):

libs/config/common/src/lib/urls.ts
```diff
@@ -36,7 +36,7 @@ const envUrls: { [key in envs]: urlMap } = {
     accountApi: 'https://account.biosimulations.dev/',
     dispatchApi: 'https://run.api.biosimulations.dev/',
     combineApi: 'https://combine.api.biosimulations.dev/',
-    simulatorsApi: 'https://api.biosimulators.dev/',
+    simulatorsApi: 'http://localhost:3333/',
```
bilalshaikh42: This needs to be changed back. Is any of the code still using this for the URLs instead of the Endpoints?

jonrkarr: That was a mistake. The code does use the URLs, mainly for hyperlinks. This was only partially refactored.

jonrkarr:
This follows #3164. A compound unique constraint would no longer be needed because its replaced by a unique constraint on a single attribute whose value is already a compound of id and version.

bilalshaikh42:
Closing based on decision to close #3164

PR #3352, opened 2021-10-21
chore(combine-service): restored dynamic reference to logging schemas in ApiCurio

jonrkarr:
This just reorders the schemas in the spec file

### DIFF
--- a/apps/combine-service/src/spec/spec.yml
+++ b/apps/combine-service/src/spec/spec.yml
@@ -3619,6 +3619,176 @@ components:
       example:
         _type: Environment
         variables: []
+    GetCombineArchiveMetadataFileOrUrl:
+      description: "File or a URL for getting the information in an [OMEX Metadata](https://sys-bio.github.io/libOmexMeta/)\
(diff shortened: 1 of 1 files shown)


bilalshaikh42:
Should this now trigger a rebuild of the combine-api-client library? Would be good to confirm that works correctly

bilalshaikh42:
> Should this now trigger a rebuild of the combine-api-client library? Would be good to confirm that works correctly

Does seem like it ran, but there were no changes to commit

jonrkarr:
There shouldn't be any changes to commit. The schemas are just in a different order in the YAML file to be in sync with what ApiCurio outputs.

The change here is really in ApiCurio. In ApiCurio, the CombineArchiveLog schema is intended to be linked to the the OpenAPI spec for the API (`https://api.biosimulations.org/openapi.json#/components/schemas/...`) to avoid defining schemas twice. When I edited the spec manually to separate the types of each response, I started from the de-referenced schema and forgot to maintain this reference. This restores that reference.

PR #3361, opened 2021-10-22
feat(api): added database model and validation endpoint for simulation run logs

### LINKED ISSUE #2048: Add database schema and validation for simulation logs
An endpoint analogous to the simulator/validate endpoint would be useful. Then we can use this to create a test case within the simulators test suite which checks that simulation tools produce valid logs.

### LINKED ISSUE #3370: Apply `BiosimulationsValidationExceptionFactory` to simulators API
I believe this should be applied to the simulators API

```typescript
app.useGlobalPipes(
    new ValidationPipe({
      exceptionFactory: BiosimulationsValidationExceptionFactory,
      transform: true,
      transformOptions: {
        enableImplicitConversion: true,
      },
    }),
  );
```

jonrkarr:
closes #2048 
closes #3370

@bilalshaikh42 a few issues still need to be addressed:

## Log validation errors are reported with code 500
Example:
```
{
  "error": [
    {
      "status": "500",
      "title": "Database CastError",
      "detail": "Cast to Number failed for value \"x\" (type string) at path \"duration\"",
      "source": {
        "pointer": "/duration"
      }
    }
  ]
}
```

## Delete all simulators executes successfully, but raises a 500 error
```
{
  "error": [
    {
      "status": "500",
      "title": "Operation Failed",
      "detail": "\"Operation Failed\""
    }
  ]
}
```

## The simulators API appears to be missing similar validation handling to the API
The simulators API doesn't have this:
```
app.useGlobalPipes(
    new ValidationPipe({
      exceptionFactory: BiosimulationsValidationExceptionFactory,
      transform: true,
      transformOptions: {
        enableImplicitConversion: true,
      },
    }),
  );
```

### DIFF
--- a/apps/simulators-api/src/simulators/simulators.service.ts
+++ b/apps/simulators-api/src/simulators/simulators.service.ts
@@ -72,19 +72,7 @@ export class SimulatorsService {
   public async new(doc: APISimulator): Promise<Simulator> {
     const sim = new this.simulator(doc);
     let res: Simulator;
-    try {
-      res = (await sim.save()).toJSON({ versionKey: false }) as Simulator;
-    } catch (e) {
-      if (e.name == 'MongoError' && e.code == 11000) {
-        throw new ConflictException(
-          `Simulator with id: ${e?.keyValue?.id}, version: ${e?.keyValue?.version} already exists`,
-        );
-      } else {
-        // Will be caught by other filters
-        console.log(e);
-        throw e;
-      }
-    }
+    res = (await sim.save()).toJSON({ versionKey: false }) as Simulator;
     return res;
   }
 
--- a/apps/simulators-api/src/simulators/simulators.service.ts
+++ b/apps/simulators-api/src/simulators/simulators.service.ts
@@ -1,6 +1,5 @@
 import {
   Injectable,
-  ConflictException,
   NotFoundException,
   InternalServerErrorException,
 } from '@nestjs/common';
@@ -71,8 +70,7 @@ export class SimulatorsService {
 
   public async new(doc: APISimulator): Promise<Simulator> {
     const sim = new this.simulator(doc);
-    let res: Simulator;
-    res = (await sim.save()).toJSON({ versionKey: false }) as Simulator;
+    const res: Simulator = (await sim.save()).toJSON({ versionKey: false }) as Simulator;
     return res;
   }
 
--- a/apps/api/src/files/files.controller.ts
+++ b/apps/api/src/files/files.controller.ts
@@ -137,7 +137,7 @@ export class FilesController {
   //@Post(':runId')
   @permissions('write:Files')
   public async createSimulationFiles(
-    @Param() runId: string,
+    @Param('runId') runId: string,
     @Body() files: any[],
   ): Promise<void> {}
 
--- a/apps/api/src/logs/logs.controller.spec.ts
+++ b/apps/api/src/logs/logs.controller.spec.ts
@@ -39,7 +39,7 @@ describe('LogsController', () => {
   it(' should get logs from service', async () => {
     const getSpy = jest.spyOn(mockService, 'getLog').mockReturnValue('test');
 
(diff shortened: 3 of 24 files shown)


bilalshaikh42:
> ## Log validation errors are reported with code 500
> Example:
> 
> ```
> {
>   "error": [
>     {
>       "status": "500",
>       "title": "Database CastError",
>       "detail": "Cast to Number failed for value \"x\" (type string) at path \"duration\"",
>       "source": {
>         "pointer": "/duration"
>       }
>     }
>   ]
> }
> ```

This is intentional. See https://github.com/biosimulations/biosimulations/tree/dev/libs/shared/exceptions/filters/src/lib/filters/Mongo
to address this, there should be validation applied at the controller level. Alternatively, the controller can catch the error and return the appropriate 400 error

> The simulators API appears to be missing similar validation handling to the API
> The simulators API doesn't have this:
> 
> app.useGlobalPipes(
>     new ValidationPipe({
>       exceptionFactory: BiosimulationsValidationExceptionFactory,
>       transform: true,
>       transformOptions: {
>         enableImplicitConversion: true,
>       },
>     }),
>   );

This should fix the above issue. I had not included this, because when I created the validation classes/exception we did not have the class-validation annotation enabled for the simulator-api. To have this work well, we need to add class-validation such as described here: https://docs.nestjs.com/techniques/validation#auto-validation



> Delete all simulators executes successfully, but raises a 500 error
> {
>   "error": [
>     {
>       "status": "500",
>       "title": "Operation Failed",
>       "detail": "\"Operation Failed\""
>     }
>   ]
> }


This is a bug that I'll create a separate issue for

apps/simulators-api/src/simulators/simulators.service.ts
```diff
@@ -72,19 +72,7 @@ export class SimulatorsService {
   public async new(doc: APISimulator): Promise<Simulator> {
     const sim = new this.simulator(doc);
     let res: Simulator;
-    try {
-      res = (await sim.save()).toJSON({ versionKey: false }) as Simulator;
-    } catch (e) {
-      if (e.name == 'MongoError' && e.code == 11000) {
```
bilalshaikh42, line 78: Good catch, this is redundant with the app level filter now

bilalshaikh42 (approved):

bilalshaikh42:
lgtm 

@jonrkarr do you want me to adress the issues you listed?

jonrkarr:
> do you want me to adress the issues you listed?

That would be great. I'm not sure exactly what you have in mind.

apps/simulators-api/src/simulators/simulators.service.ts
jonrkarr: I figured this was old. Also, in my experience the error name is (now) `MongoServerError`.

bilalshaikh42:
> ## remove sha256 prefix from hash
> https://github.com/biosimulations/biosimulations/blob/b01630a5e24144ec599c185d1cb5cb5463085c8e/libs/datamodel/api/src/lib/simulators/image.ts#L25-L29
> 
> ###### This comment was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in [b01630a](https://github.com/biosimulations/biosimulations/commit/b01630a5e24144ec599c185d1cb5cb5463085c8e) in #3361. cc @biosimulations.

@jonrkarr  Would it make sense to remove the 'sha256:' prefix from the images? that would allow us to use one of the validators.

jonrkarr:
I contemplated that. I kept that in there for consistency with Docker, GitHub, OCI, etc. 

Is it possible to create a validator?

bilalshaikh42:
> I contemplated that. I kept that in there for consistency with Docker, GitHub, OCI, etc.
> 
> Is it possible to create a validator?

That makes sense.  Yup, just a few extra steps. I'll add one in

jonrkarr:
Fixing issues in #3367

jonrkarr:
Manually checked stalled build step.

PR #3418, opened 2021-10-29
feat: added endpoints for getting summaries of projects

jonrkarr:


bilalshaikh42:
what information is included in the summary? can we just add that info to the projects object?

jonrkarr:
This is information derived from other collections:
- SED tasks
- SED outputs and Vega visualizations
- Top-level metadata

It could be stored in MongoDB. Instead, I'm caching it.

PR #3444, opened 2021-10-31
refactor(dispatch): added re-trying for similar algorithms

### LINKED ISSUE #3434: Add retrying for getting similar algorithms from COMBINE service
This sometimes fails when the service startups up because the computation sometimes times out before the cache is created

jonrkarr:
Failures are due to long (few seconds) burn in time. Due to server-side caching in Python which stores the successful part of the computation, this is alleviating by simply retrying the request.

closes #3434


jonrkarr:
Tests are failing due to timeouts. This is unrelated to this PR, and appearing in others.

PR #3452, opened 2021-11-02
refactor(api): finishing up cascading deletion

jonrkarr:
- [x] delete files from S3
- [x] test and debug deletion
- [x] Move hardcoded S3 and HSDS paths to `Endpoints`


jonrkarr:
Merged manually

PR #3457, opened 2021-11-02
refactor(dispatch,dispatch-service): added post-processing to stdout log

### LINKED ISSUE #3451: Simplify tracking of post-processing failures
Currently, post-processing failures can only be distinguished from the status reason. It would be helpful to create a distinct status code for this failure.

- [ ] Create distinct status code
- [ ] Update `apps/dispatch/src/app/components/simulations/view/simulation-log/simulation-log.component.ts` to use this code to format post-processing errors

jonrkarr:
- [x] Organized console formatting commands
- [x] Cleaned up log displaying
- [x] Added post-processing status to combined stdout/err log -- closes #3451 
- [x] Enabled log PUT method
- [x] Improve handling of case when YAML log file doesn't exist -- treat this as an optional simulator feature rather than an error
- [x] Handle errors in logs by posting stdout
- [x] Improve log messages
- [x] Encapsulate arguments in SBATCH script in quotes

### DIFF
--- a/apps/dispatch-service/src/app/submission/complete.processor.ts
+++ b/apps/dispatch-service/src/app/submission/complete.processor.ts
@@ -144,10 +144,10 @@ export class CompleteProcessor {
     const extraStdLog = (
       ''
       + '\n'
-      + `\n${ConsoleFormatting.cyan}======================= Post-processing simulation run ======================${ConsoleFormatting.noColor}`
+      + `\n${ConsoleFormatting.cyan}=========================================== Post-processing simulation run ==========================================${ConsoleFormatting.noColor}`
       + `\n${statusColor}statusReason${statusEndColor}`
       + '\n'
-      + `\n${ConsoleFormatting.cyan}============ Run complete. Thank you for using runBioSimulations! ===========${ConsoleFormatting.noColor}`
+      + `\n${ConsoleFormatting.cyan}================================ Run complete. Thank you for using runBioSimulations! ===============================${ConsoleFormatting.noColor}`
     );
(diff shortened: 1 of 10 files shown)


apps/dispatch-service/src/app/submission/complete.processor.ts
```diff
@@ -152,21 +153,49 @@ export class CompleteProcessor {
...
+
+    const extraStdLog = (
+      ''
+      + '\n'
+      + `\n${ConsoleFormatting.cyan}======================= Post-processing simulation run ======================${ConsoleFormatting.noColor}`
+      + `\n${statusColor}statusReason${statusEndColor}`
+      + '\n'
+      + `\n${ConsoleFormatting.cyan}============ Run complete. Thank you for using runBioSimulations! ===========${ConsoleFormatting.noColor}`
+    );
+    this.logService.createLog(id, extraStdLog, true)
```
bilalshaikh42: This does not seem right. I think this was already an issue before that just got copied over, but the then does not catch errors, but successful updates

bilalshaikh42:
What is the intention of the put method? Is this being called by the dispatch service?

bilalshaikh42 (approved):
LGTM

apps/dispatch-service/src/app/submission/complete.processor.ts
jonrkarr: Errors are handled by `const logPostSucceeded = logProcessingResult.status === 'fulfilled' && logProcessingResult?.value;` and the second call to `createLog` (`this.logService.createLog(id, !logPostSucceeded, extraStdLog, true)`)

PR #3486, opened 2021-11-03
fix(dispatch,api,dispatch-service): fixed data model for exceptions in simulation run logs

### LINKED ISSUE #3482: [Bug]: Dispatch service does not produce logs that can be properly validated by the api
### Contact Details

_No response_

### Location

BioSimulations API

### What happened?

When running a simulation with an invalid omex archive, the job fails as expected. However, the dispatch service's upload of the log also fails due to the following validation error

```
               "type": "ValueError"
              },
              "property": "exception",
              "children": [
                {
                  "target": {
                    "message": "`/root/input.omex` is not a file.",
                    "type": "ValueError"
                  },
                  "property": "category",
                  "children": [],
                  "constraints": {
                    "isString": "category must be a string",
                    "isNotEmpty": "category should not be empty"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

### What made it happen?

Running a simulation that fails

### Version

Development (Deployed on a .dev domain)

### What browsers are you seeing the problem on?

_No response_

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jonrkarr:
Closes #3482. Also reverts temporary fix in #3484.

### DIFF
--- a/apps/api/src/logs/logs.model.ts
+++ b/apps/api/src/logs/logs.model.ts
@@ -30,7 +30,7 @@ import { SimulationRunModel } from '../simulation-run/simulation-run.model';
 })
 export class Exception implements IException {
   @Prop({ type: String, required: true, default: undefined })
-  category!: string;
+  type!: string;
 
   @Prop({ type: String, required: true, default: undefined })
   message!: string;
--- a/apps/combine-service/src/spec/spec.yml
+++ b/apps/combine-service/src/spec/spec.yml
@@ -3815,11 +3815,11 @@ components:
             $ref: '#/components/schemas/SedDocumentLog'
     Exception:
       required:
-      - category
+      - type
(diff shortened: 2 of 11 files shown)


bilalshaikh42:
@jonrkarr can you include a  `BREAKING CHANGE: <description>` footer to one of the commit messages with a note about  changes to log schema? Not just the category => type change, but any of the change that were a part of the validation of the logs/handling of errors

bilalshaikh42:
We would also need to regenerate the combine-api client

jonrkarr:
I updated the COMBINE API client. Its not really a breaking change because this is what simulators have been outputting all along. This corrects the apps so they correctly display errors from simulators.

Issue #3488, opened 2021-11-03
Set Up nx build targets for COMBINE-API


bilalshaikh42:
@jonrkarr could you add the config for building the combine-API docker images to https://github.com/biosimulations/Biosimulations/blob/08674bfaeb3cf068d7e5e74bbc6a7154671b93c6/apps/combine-service/project.json

I have added the config to the typescript apps and most of the python app  in #3492. Just need to verify if there are any build-args that need to be included

jonrkarr:
Not sure what the syntax is. 

This is the basic command to build the image
```
docker build \
  --build-arg app=combine-service \
  -f DockerfilePython \
  .
```

Its probably easiest to run the tests in the Docker image. This avoids needed to build the environment twice.
```
docker run \
            --rm \
            --entrypoint bash \
            --mount type=bind,source=$(pwd)/apps/combine-service,target=/app/apps/combine-service \
            ghcr.io/biosimulations/combine-service:latest \
            -c "
              pipenv install --dev --system --deploy
              pip uninstall -y matplotlib
              pip install matplotlib==3.2.0
              mv src src-ignore
              export PYTHONPATH=apps/combine-service
              /bin/bash /xvfb-startup.sh python -m pytest --verbose --cov apps/combine-service/src/ --cov-report=xml apps/combine-service/tests/
            "
```

bilalshaikh42:
Sorry, I had not yet pushed the changes I was referencing.  #3492 adds a nx  builder to create and push docker images for each app. Could you confirm the python docker image builds correctly?

jonrkarr:
Will be fixed in #3496.

PR #3492, opened 2021-11-03
build: add nx tooling support for building dockerfiles. Add definitions for backend apps

bilalshaikh42:
This allows for the apps docker images to be built by using the `nx docker [target]` command, (e.g `nx docker api`) to create a docker image and push to ghcr.  The usage parallels that of `nx serve [target]` etc.


bilalshaikh42:
@jonrkarr This is the config that needs to be confirmed from #3488

Issue #3628, opened 2021-11-16
Requests from machine accounts no longer include user information (no AuthToken)


jonrkarr:
Tokens are need to track ownership of projects. This information is no longer included in requests. For machine accounts, the `user` object is now empty.

bilalshaikh42:
I am surprised that they were previously included. Machine apps inherently are not users, so I would expect the user object to be empty. 

The token for the machine apps looks like this to me

```
{
  "iss": "https://crbm.auth0.com/",
  "sub": "5BRH4SlefREz7YRRuUeEo9tNf9HP9k2f@clients",
  "aud": "api.biosimulations.org",
  "iat": 1637085609,
  "exp": 1637172009,
  "azp": "5BRH4SlefREz7YRRuUeEo9tNf9HP9k2f",
  "scope": "read:Projects update:Projects create:Projects",
  "gty": "client-credentials",
  "permissions": [
    "read:Projects",
    "update:Projects",
    "create:Projects"
  ]
}
```

A user identity token would look similar but potentially contain additional metadata.  The "sub" claim provides the unique identity of the user and/or machine client. 

Are you specifically talking about the  `User` decorator in auth library?  If this was working previously, then it was not intentional. However, it can be adapted to provide a User or Client depending on what makes the call.

jonrkarr:
I'm referring to the `@Req` decorator and the `user` attribute of `Request`. This previously had tokens for both users and machine accounts. Now it only has tokens for users.

import { Req } from '@nestjs/common';
import { Request } from 'express';

bilalshaikh42:
Can you point me to the code where this is happening?

jonrkarr:
This is marked in the comment in https://github.com/biosimulations/biosimulations/pull/3629

https://github.com/biosimulations/biosimulations/blob/d3258aaca6342f324e4e68aca59f9dcabc286bf5/apps/api/src/simulation-run/simulation-run.controller.ts#L232-L237

This is how I'm submitting requests for simulation runs with requests to publish the run upon its successful completion:
```python
import json
import requests
import subprocess

client_id = '...'
client_secret = '...'

response2 = requests.post('https://crbm.auth0.com/oauth/token', json={
    'client_id': client_id,
    'client_secret': client_secret,
    'audience': 'api.biosimulations.org',
    "grant_type": "client_credentials",
})
response2.raise_for_status()
response_data = response2.json()
[...]

PR #3629, opened 2021-11-16
feat(api): restrict permission to externally validate simulation runs to model repositories and super users

### LINKED ISSUE #3612: Only allow repositories to include projectIds in simulation run requests
Add logic add to controller for simulation run POST endpoint

Check `type` of the login token

jonrkarr:
Closes #3612

~~@bilalshaikh42 this is ready to merge, except this needs to correctly check permissions for machine accounts.~~

### DIFF
--- a/apps/api/src/simulation-run/simulation-run.controller.ts
+++ b/apps/api/src/simulation-run/simulation-run.controller.ts
@@ -229,7 +229,7 @@ export class SimulationRunController {
   private checkPublishProjectPermission(user: AuthToken, projectId?: string): void {
     if (
       projectId
-      && !user?.permissions?.includes(scopes.simulationRuns.externallyValidate.id) // TODO: debug for machine accounts
+      && !user?.permissions?.includes(scopes.simulationRuns.externallyValidate.id)
     ) {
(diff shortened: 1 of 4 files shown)


apps/api/src/simulation-run/simulation-run.controller.ts
```diff
@@ -217,33 +219,38 @@ export class SimulationRunController {
...
     };
     const sim = await this.dispatchQueue.add(message);
 
     return response;
   }
 
+  private checkPublishProjectPermission(user: AuthToken, projectId?: string): void {
+    if (
+      projectId
+      && !user?.permissions?.includes(scopes.simulationRuns.externallyValidate.id) // TODO: debug for machine accounts
```
jonrkarr: @bilalshaikh42 This is the point where machine account information is no longer available

bilalshaikh42: I am not able to replicate this. On #3628 I see an issue with the code you are using. Does changing that fix it?

bilalshaikh42 (approved):

PR #3772, opened 2021-12-09
perf(api): extract the combine archive directly on s3 via streams

### LINKED ISSUE #3094: Move combine archive extraction and uploading to s3 from hpc to dispatch-service
The combine archive is unzipped on the HPC and then uploaded along with the rest of the output to s3 at the end of the simulations run. 
It might make sense to have the API/dispatch service handle this 
Pros
1. can get information about size directly. Currently are using additional HTTP calls to make head requests to get size 
2. can get URL information directly instead of hardcoding it throughout app
3. can keep track of the URL for the archive itself (alternatively, just name all the archives as input.omex)
4. One less step on slurm to track/monitor
5. Speedup? 
Cons:
1. harder than just using the sync CLI tool 
2. Need to coordinate uploading the outputs 
3. Keep track of when its complete

bilalshaikh42:
uses streams to directly extract the combine archive onto s3 as soon as the archive is uploaded.
This removes the "extract" step from the slurm workflow, and is significantly faster. This also
opens up the possibility of the combine-api responding to an event to access the files that it needs
rather than waiting for the post processing to send it the files from the simulation run.

closes #3094


bilalshaikh42:
@jonrkarr it turned out to be much easier to extract the archive directly on s3 than using the HPC. This opens up the possibility of doing a lot of the post-processing before/while the simulation is running rather than at the end. Would the combine API be able to now use the saved files rather than download and extract the combine archive for each request as u mentioned in #3094?

PR #3854, opened 2021-12-21
feat(api): minor polishing

### LINKED ISSUE #3759: check deletion worked and all needed files were deleted
https://github.com/biosimulations/biosimulations/blob/751c39ee61238a3e784133de1cf78e0dea291d9b/libs/shared/storage/src/lib/simulation-storage.service.ts#L36-L41

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 751c39ee61238a3e784133de1cf78e0dea291d9b when #3753 was merged. cc @biosimulations.

### LINKED ISSUE #3777: clean lint warning below and update warning to errors
https://github.com/biosimulations/biosimulations/blob/1ad248aea0512204047ed8d5b75fcbe058253565/apps/api/src/simulation-run/simulation-run.service.ts#L550-L553

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 1ad248aea0512204047ed8d5b75fcbe058253565 when #3772 was merged. cc @biosimulations.

### LINKED ISSUE #3778: remove the redundancy between the various s3 paths
https://github.com/biosimulations/biosimulations/blob/1ad248aea0512204047ed8d5b75fcbe058253565/libs/config/common/src/lib/endpoints.ts#L454-L459

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in 1ad248aea0512204047ed8d5b75fcbe058253565 when #3772 was merged. cc @biosimulations.

jonrkarr:
- closes #3759
- closes #3777
- closes #3778
  - [x] Organize S3 paths to shared storage library (`libs/shared/storage`)
  - [x] Organize HSDS paths to hsds client library (`libs/hsds/client`)
  - [x] Organize app routes to separate class (`libs/config/common/src/lib/app-routes.ts`)
  - [x] Organize resource URIs (identifiers.org) to separate class (`libs/config/common/src/lib/resource-identifiers.ts`)
  - [x] test
- [x] Remove duplicate files in `libs/combine-api/angular-client/`
- [x] Add project id, owner to `CompleteJob` for failures


libs/shared/storage/src/lib/simulation-storage.service.ts
```diff
@@ -36,7 +37,17 @@ export class SimulationStorageService {
     );
     const output = await this.storage.deleteObject(fileObject);
 
-    // TODO check deletion worked
+    await this.storage.getObject(fileObject)
+      .then((file: AWS.S3.GetObjectOutput): void => {
+        if (!file.DeleteMarker) {
```
bilalshaikh42: @jonrkarr  Were you able to get this to work? 
As far as I could tell, this approach does not work and will always throw an error. The delete marker does not get set on files when deleted this way. I believe it has something to do with the way versioning is used on the s3 bucket

libs/shared/storage/src/lib/simulation-storage.service.ts
```diff
@@ -69,8 +80,19 @@ export class SimulationStorageService {
...
+    await this.storage.deleteObject(fileObject);
+
+    await this.storage.getObject(fileObject)
+      .then((file: AWS.S3.GetObjectOutput): void => {
+        if (!file.DeleteMarker) {
+          throw new InternalServerErrorException(`COMBINE archive could not be deleted for simulation run '{runId}'.`);
+        }
+      })
+      .catch((error: any): void => {
+        if (error.status !== HttpStatus.NOT_FOUND && error.statusCode !== HttpStatus.NOT_FOUND) {
```
bilalshaikh42: This won't ever catch since S3 will not throw an error. Instead, the s3GetObjectOutput type will contain an error that has information about the error that occurred. Similarly, there won't be standard http status codes to test for here

libs/config/common/src/lib/endpoints.ts
```diff
@@ -479,24 +476,26 @@ export class Endpoints {
...
+    return `${this.simulationRunsS3Path}/${runId}${subPath}`;
   }
 
   /**
    * Create a path for the COMBINE/OMEX archive of a simulation run in an S3 bucket
    * @param runId Id of the simulation run
    */
   public getSimulationRunCombineArchiveS3Path(runId: string): string {
-    return `${this.getSimulationRunS3Path(runId)}/archive.omex`;
+    return this.getSimulationRunS3Path(runId, 'archive.omex');
```
bilalshaikh42: Rather than rely on endpoints class to encode the paths here, I think it would be best to have the details/knowledge of the file organization on s3 all contained within the simulation-storage service and remove the dependency on endpoints entirely. Nothing else in the system needs to know how the s3 buckets are organized once the combine archives are extracted since they all then have URLs attached to them and are saved in the database. Keeping this here just encourages some other part of the code to rely on this information, causing coupling where it does not need to be

bilalshaikh42 (changes requested):

libs/shared/storage/src/lib/simulation-storage.service.ts
jonrkarr: The combination seemed to work

libs/shared/storage/src/lib/simulation-storage.service.ts
jonrkarr: This is intended to catch files which were deleted (404 code). This seemed to work in my testing.

libs/config/common/src/lib/endpoints.ts
jonrkarr: Feel free to create a separate issue to move this.

libs/shared/storage/src/lib/simulation-storage.service.ts
bilalshaikh42: Was this on the UConn bucket? As far as I can tell from the docs, the delete marker is only set for versioned objects. If the bucket is not versioned, the file should not exist. If the tag is set, then maybe versioning is enabled on the bucket?

libs/shared/storage/src/lib/simulation-storage.service.ts
bilalshaikh42: That confuses me, since if the delete marker is set as above, then the object itself still exists, though its versions may have been deleted. This means we should never get a 404.  if we are getting a 404, then the the file should always be null

bilalshaikh42 (changes requested):

libs/config/common/src/lib/endpoints.ts
jonrkarr: The configuration for the S3 paths is shared between the frontend and backend. This has to be separated before this configuration can be organized into the shared storage library.

bilalshaikh42: yes, this is the main problem. The front end really shouldn't be aware of the s3 bucket at all, but get any files through the API.  
I will take a look at this when I return next week

jonrkarr: I separated this and organized the S3 paths into the shared storage library.

libs/shared/storage/src/lib/simulation-storage.service.ts
jonrkarr: With out bucket, after deletion, `getObject` raises an error with `statusCode` `404` and `code` `NoSuchKey`. The if statement in the catch catches these errors and re-throws all other errors.

The code was written to handle both cases when the bucket is and isn't versioned. For clarity, I removed the DeleteMarker check.

jonrkarr:
I tested deletion using the UConn S3 bucket. Here's what happens with this bucket:

- Prior to calling `deleteObject`, the output of `getObject` has `DeleteMarker` `undefined`
- `deleteObject` deletes S3 objects immediately.
- The output of `deleteObject` is an instance of `AWS.S3.DeleteObjectOutput` with value `{ DeleteMarker: true, VersionId: 'null' }`
- Following deletion, `getObject` throws an instance of `Error` with `statusCode` `404` and `code` `NoSuchKey`
    ```
    Error {
      code: 'NoSuchKey',
      region: null,
      time: 2021-12-21T18:40:17.922Z,
      requestId: 'ab87a8f8c545aef759cb',
      extendedRequestId: 'ab87a8f8c545aef759cb',
      cfId: undefined,
      statusCode: 404,
      retryable: false,
      retryDelay: 3.4178620828735573
    }
    ```
- `getObject` does not return null or an instance of `S3GetObjectOutput`

The code committed yesterday was corrected and handled the case that the object isn't deleted immediately. For clarity, I removed the code for this additional case. I also sharpened the error handling to check that the `code` of the error raised by `getObject` is [`NoSuchKey`](https://docs.aws.amazon.com/ja_jp/AWSJavaScriptSDK/v3/latest/clients/client-s3/modules/nosuchkey.html). The string `NoSuchKey` is hard-coded because this isn't defined in the AWS SDK package, although it is described in the documentation online.

jonrkarr:
I added an argument to control whether to confirm deletion or not with `getObject`. The default for this is currently `true`. I think the default can be changed to false. This confirmation seems unnecessary because it appears that deletion can also be confirmed by checking the `DeleteMarker` attribute of the output of `deleteObject`.

PR #3909, opened 2021-12-27
docs: add bullets to resources list

freiburgermsu:


bilalshaikh42:
@freiburgermsu Thanks for these! Could you make sure to use conventional commits for changes?  This helps us keep track of changes. Here is some info https://www.conventionalcommits.org/en/v1.0.0/

This commit would be `docs: add bullets` for example.

bilalshaikh42 (approved):

PR #3922, opened 2021-12-28
two suggestions

freiburgermsu:


bilalshaikh42:
The command line utils also refer to `for creating, validating, and executing simulation projects`

PR #4001, opened 2022-01-08
docs: correct cookie policy regarding required cookies

### LINKED ISSUE #4000: Cookie Policy is incorrect
The cookie policy states that the use of local storage to save simulation runs is 'functional', but it is actually 'required' 

'functional' refers to added functionality that can be opted out of such as remembering preferences or enabling additional sharing options. 
 
Since the saving of simulation runs to the the browser is a core  feature that cannot be turned off without breaking the operation of the site, this is 'a strictly required' use.  In general, anything that is 'functional' MUST respect the user opting out of functional cookies and check for consent before saving any data.

bilalshaikh42:


jonrkarr (approved):
Looks good. Thanks for catching this.

PR #4029, opened 2022-01-11
docs(combine-api): added documentation for installation and execution

bilalshaikh42:


jonrkarr (changes requested):
An accurate README would be helpful. Lets provide a more accurate description of the installation and execution processes because `Pipfile`/`Pipfile.lock` isn't sufficient. 

- [x] Either expand the instructions (more work) or add a couple notes (easier, what I recommend):
   - These instructions install most dependencies
   - See `Dockerfile` for a complete installation reference

- [x] Also, the run instructions need to be corrected. Either (a) `python -m ...` has to be run from the pipenv environment or (b) the `Pipfile` needs to be installed into the system Python environment.

PR #4040, opened 2022-01-12
perf(combine-api): improve production runner for combine-api

bilalshaikh42:


jonrkarr (approved):
I was also able to replicate the crashing. This occurs just from loading the swagger app, without querying any endpoints. Consequently, i think it's likely related to a swagger dependency. Hopefully i can narrow it down today.

bilalshaikh42:
Ill deploy `ghcr.io./biosimulations/combine-api:pr-4040` to dev and see if that helps. There might still need to be some more tweaks

PR #4042, opened 2022-01-12
perf(api, dispatch-service): switch storage backend to Google Cloud storage bucket

### LINKED ISSUE #4052: Fix code scanning alert - Clear-text logging of sensitive information
Tracking issue for:
- [ ] https://github.com/biosimulations/biosimulations/security/code-scanning/57

bilalshaikh42:


apps/dispatch-service/src/file/file.service.ts
```diff
@@ -82,47 +83,32 @@ export class FileService {
...
-                      `The size of '${
-                        file.location.path
-                      }' for simulation run '${id}' could not be retrieved: ${this.getErrorMessage(
-                        error,
-                      )}`,
+              const fileInfo = this.storage
+                .getFileInfo(id, file.location.path)
+                .pipe(
+                  map((fileInfo: FileInfo): ProjectFileInput => {
+                    // TODO do we want to throw here if we don't have a size?
```
bilalshaikh42: @jonrkarr You were throwing an error here if we were unable to get the file size. Is this something that we should do here? I am currently just setting the size to undefined if we are not able to get it, as it feels silly to fail the whole run just because one of the file sizes is missing. 
What do you think?

jonrkarr:
Not throwing an error is fine. Previously i believe the size was set to zero if it couldn't be obtained. Better would be undefined. The UI is already set up to display undefined as N/A.

bilalshaikh42:
@jonrkarr still needs to add/update tests and do a final end to end check, but everything should be good for review

PR #4086, opened 2022-01-19
perf(dispatch-service): maintain persistent ssh connections

bilalshaikh42:
refactored the ssh service to maintain a persistent ssh connection rather than create and close the connection for each command
also added event listners to maintain the connection and reconnect if there is an error
this should speed up the commands as well as reduce errors due to too many connections.


jonrkarr:
Looks fine to me. 

I think the log and output size could be obtained with SSH from the S3 bucket.

PR #4093, opened 2022-01-20
perf(dispatch-service): switched log and output retrieval to use S3 rather than SSH

### LINKED ISSUE #4088: Get logs and output size from S3 rather than SSH

jonrkarr:
Closes #4088

Avoids another SSH interaction with UConn HPC.


bilalshaikh42:
Let's hold off on this until #4087 which is reorganizing all the steps into queues

bilalshaikh42:
Covered in #4087, along with simplifying log processing

Issue #4078, opened 2022-01-19
Add and ensure permissions for service accounts to use multi-part uploads


bilalshaikh42:
Running the sumbit script casues the following error for the RVA simulation. Disabled this particular one allows remaining simulations to proceed.

```
Traceback (most recent call last):
  File "/home/bilal/Projects/work/Biosimulations/tools/submit-example-simulation-runs", line 383, in <module>
    main(runbiosimulations_deployment=args.runbiosimulations_deployment,
  File "/home/bilal/Projects/work/Biosimulations/tools/submit-example-simulation-runs", line 198, in main
    raise RuntimeError('Simulation could not be run on {}: {}'.format(
RuntimeError: Simulation could not be run on https://api.biosimulations.dev/runs: [{'status': '500', 'title': 'An error occurred in uploading the COMBINE archive for the simulation run: Access denied..', 'meta': {'err': {}, 'time': 1642555208971, 'url': '/runs'}}]
```

Im confused that this seems to be happening when saving the combine archive to the cloud bucket. There might some particular filename causing this. I have no idea how this can cause a permissions issue. This might be an issue with downloading the archive that  the error handling is misattributing to uploading the archive

## Todo
- [x] Adjust bucket settings to allow larger files
- [x] Re-enable RBA example simulation in `tools/example-projects.json`

- [ ]  Enable and test permissions for dev service account 
- [ ] Enable and test permissions for prod service account

jonrkarr:
This archive is larger than the others (12 MB). Is there a file size limit configuration on the bucket?

jonrkarr:
I can replicate this with boto3, other files, other S3 keys, and other buckets.

This must be due to configuration of the bucket. There appears to be a limit at 8388608 bytes = 8 MB.

Maybe this relates to `content-length-range`. See https://cloud.google.com/storage/docs/authentication/signatures#:%7E:text=Content%20Length%20Range

Here's some Python code to easily replicate this.

```python
import boto3
import os
import tempfile

endpoint_url = 'https://storage.googleapis.com'
bucket_name = 'biosim-temp-dev'
aws_access_key_id = '...'
aws_secret_access_key = '...'

s3 = boto3.resource(
    's3',
    endpoint_url=endpoint_url,
    aws_access_key_id=aws_access_key_id,
    aws_secret_access_key=aws_secret_access_key,
    verify=False,
)
bucket = s3.Bucket(bucket_name)

[...]

PR #4095, opened 2022-01-20
perf(api): switched to streaming downloaded files to storage bucket

jonrkarr:
Tried to explore #4078, but haven't figured it out.

- Switched to streaming files downloaded from URLs to the storage bucket. Follows https://dev.to/vikasgarghb/streaming-files-to-s3-using-axios-h32.
- Improved error messages and comments
- Updated object deletion in `apps/combine-api/src/s3.py` to work with Google object storage.

### DIFF
--- a/libs/shared/storage/src/lib/shared-storage.service.ts
+++ b/libs/shared/storage/src/lib/shared-storage.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, HttpStatus, Logger, RequestTimeoutException, InternalServerErrorException } from '@nestjs/common';
+import { Injectable, HttpStatus, Logger } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { InjectS3, S3 } from 'nestjs-s3';
 import * as AWS from 'aws-sdk';
@@ -6,6 +6,7 @@ import { AWSError } from 'aws-sdk';
 import { Readable, PassThrough } from 'stream';
 import unzipper, { File } from 'unzipper';
 import { promiseRetry } from './promise-retry/promise-retry';
+import { BiosimulationsException } from '@biosimulations/shared/exceptions';
 
 @Injectable()
 export class SharedStorageService {
@@ -161,7 +162,11 @@ export class SharedStorageService {
         this.logger.error(
           `Timeout when uploading '${key}' to storage in ${this.S3_UPLOAD_TIMEOUT_TIME} ms.`,
         );
-        throw new RequestTimeoutException('Timeout when uploading file to storage.');
+        throw new BiosimulationsException(
+          HttpStatus.REQUEST_TIMEOUT,
+          'File could not be saved',
+          'Timeout when uploading file to storage.',
+        );
       } else {
         const details =
           err instanceof Error && err.message
@@ -173,7 +178,11 @@ export class SharedStorageService {
           err instanceof Error && err.message
             ? err?.message
             : 'Error when uploading file to storage.';
-        throw new InternalServerErrorException(message);
+        throw new BiosimulationsException(
+          HttpStatus.INTERNAL_SERVER_ERROR,
+          'File could not be saved',
+          message,          
+        );
       }
     }
   }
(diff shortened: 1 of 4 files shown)


bilalshaikh42:
> * Switched to streaming files downloaded from URLs to the storage bucket. Follows https://dev.to/vikasgarghb/streaming-files-to-s3-using-axios-h32.

I believe this is already being done since #3772. Do you know how this differs? I am not too familiar with nodejs streams, but the way I set it up, I thought was already using streams.

libs/shared/storage/src/lib/shared-storage.service.ts
```diff
@@ -126,19 +127,24 @@ export class SharedStorageService {
...
+    key: string,
     data: Buffer | Readable,
     isPrivate = false,
   ): Promise<AWS.S3.ManagedUpload.SendData> {
     const acl = isPrivate ? 'private' : 'public-read';
     const request: AWS.S3.PutObjectRequest = {
-      Key: id,
-      Body: data,
+      Key: key,
+      Body: isReadableStream(data)
```
bilalshaikh42, line 138: Doesn't this set the body to be a boolean?

libs/shared/storage/src/lib/shared-storage.service.ts
```diff
@@ -153,28 +159,28 @@ export class SharedStorageService {
...
-            : `Error when uploading ${id} to storage.`;
+            : `Error when uploading ${key} to storage.`;
         this.logger.error(details);
 
         const message =
           err instanceof Error && err.message
             ? err?.message
             : 'Error when uploading file to storage.';
-        throw new Error(message);
+        throw new InternalServerErrorException(message);
```
bilalshaikh42: I don't think we should use http exceptions outside of the controller/api and the handlers are designed to turn these into specific responses. For errors that we catch, we can use BiosimulationsError so that we have better control of the fields and can make better error messages

libs/shared/storage/src/lib/shared-storage.service.ts
```diff
@@ -126,19 +127,24 @@ export class SharedStorageService {
...
+      Body: isReadableStream(data) 
+        ? new PassThrough() 
+        : data,
       Bucket: this.BUCKET,
       ACL: acl,
     };
 
     const call = this.s3.upload(request).promise();
+    if (isReadableStream(data)) {      
+      (data as Readable).pipe(request.Body as PassThrough);
```
bilalshaikh42, line 147: Are these casts needed? If so, then the typegaurd doesn't seem to working, which is confusing since it seems correct to me

bilalshaikh42 (changes requested):

libs/shared/storage/src/lib/shared-storage.service.ts
jonrkarr: The value of the body key continue onto the next two lines:
```
Body: isReadableStream(data) 
        ? new PassThrough() 
        : data,
```

libs/shared/storage/src/lib/shared-storage.service.ts
jonrkarr: That didn't work for me. The casts were needed.

libs/shared/storage/src/lib/shared-storage.service.ts
jonrkarr: I tried to make the error more informative than just generic `Error`.

Scopes prevent `BiosimulationsError` from being imported here.

jonrkarr:
> I believe this is already being done since #3772. Do you know how this differs? I am not too familiar with nodejs streams, but the way I set it up, I thought was already using streams.

I'm not familiar either. The data type was stream. The S3 library seemed to handle it previously. Others suggested the method I introduced. This was me trying to figure out the Access Denied issue. I tried a few things until I convinced myself its not an issue with our TypeScript code.

bilalshaikh42:
> > I believe this is already being done since #3772. Do you know how this differs? I am not too familiar with nodejs streams, but the way I set it up, I thought was already using streams.
> 
> I'm not familiar either. The data type was stream. The S3 library seemed to handle it previously. Others suggested the method I introduced.

Instead of passing in the stream as an object, now we are piping it in? Is that the difference?

libs/shared/storage/src/lib/shared-storage.service.ts
jonrkarr: `BiosimulationsException` should be used here. I changed errors to `BiosimulationsException`.

jonrkarr:
> Instead of passing in the stream as an object, now we are piping it in? Is that the difference?

Yes. I don't know how much of a difference this makes. It doesn't address the issue I was trying to fix. But I kept it because it seemed like an improvement.

bilalshaikh42 (approved):

PR #4096, opened 2022-01-20
feat(api): standardized titles of error messages

### LINKED ISSUE #4017: Error handling repeats title and details
For example a 404: 

```JSON
{"error":
[{"status":"404",
"title":"Results could not be found for simulation run 'xxx'.",
"detail":"\"Results could not be found for simulation run 'xxx'.\""}]
```
The title should be 'Not Found'

jonrkarr:
Closes #4017 

Standardized titles using the `http-status-codes` package.

@bilalshaikh42 the function below seems to be unused. Please confirm this can be deleted.

### DIFF
--- a/libs/shared/exceptions/exceptions/src/lib/exception.ts
+++ b/libs/shared/exceptions/exceptions/src/lib/exception.ts
@@ -4,6 +4,7 @@ import {
   ErrorSourceObject,
 } from '@biosimulations/datamodel/api';
 import { HttpException } from '@nestjs/common';
+import { StatusCodes, ReasonPhrases } from 'http-status-codes';
 
 // TODO refactor to be able to hold multiple errors
 export class BiosimulationsException extends Error {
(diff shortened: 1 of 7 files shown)


libs/shared/exceptions/exceptions/src/lib/exception.ts
```diff
@@ -51,35 +63,22 @@ export class BiosimulationsException extends Error {
...
   getStatus(): number {
     return this.status;
   }
+
   getError(): ErrorObject {
     return this.errorObject;
   }
-
-  // TODO write test
-  static fromHTTP(exception: HttpException): BiosimulationsException {
```
jonrkarr, line 67: @bilalshaikh42 this isn't used anywhere. Please confirm this can be deleted.

bilalshaikh42: This was meant to be added to the error handlers for the HTTP exception filter to provide a consistent error object. If it is not used that would be the cause of #4017

bilalshaikh42: With the rest of the changes, this seems to be redundant now.

bilalshaikh42 (approved):

libs/shared/exceptions/exceptions/src/lib/exception.ts
jonrkarr: That method was never hooked up to anything. The HTTP errors and BioSimulationsExceptions seem to be separate. I decided to standardize the titles of both by using the numeric codes to look up standardized titles.

PR #4104, opened 2022-01-31
refactor: Small MatSnackBar/Clipboard refactor

jmdetloff:
Creates ClipboardService, which encapsulates writing to the clipboard and then displaying a styled MatSnackBar.

N/A

The copy to clipboard SnackBar previously included an OK button in some cases. I have removed this for consistency, and because the SnackBar isn't onscreen long enough to tap the OK button anyways.

I manually tested copying from the browse component, the breadcrumb buttons, and the view simulator component in chrome. Copying still works and the SnackBar still appears. I also wrote a test for the ClipboardService and confirmed it passes, along with the rest of the project's tests.

**Additional information**
This is a very small change I made this week while ramping up on Angular and exploring the codebase. It isn't necessary for any behavior reason, it just removes a bit of repetition in the code. I can archive it and call it practice if right now isn't a good time to pull in style tweaks.

I did make a couple assumptions that should be confirmed by reviewer:
1. That shared/angular is intended to hold services and components that are shared among the angular libraries. This is where I placed the new ClipboardService.
2. That by importing a module A into a module B it is guaranteed that module A's providers will be available for dependency injection for components declared from module B or services provided from Module B. These relationships are new to me so I want to double check them.


bilalshaikh42 (approved):
This looks great! Thank you

bilalshaikh42:
@all-contributors please add @jmdetloff for code

jonrkarr:
@jmdetloff thanks for the contribution!

Please forgive the code isn't as streamlined as ideal. We're often prioritizing integrating domain content (largely implemented in other repositories) and trying to prototype the UI quickly.

PR #4108, opened 2022-02-01
docs: add glossary for common terms

bilalshaikh42:


jonrkarr:
Embedding the entire glossary in every page seems a bit repetitive. Would a more prominent top-level menu item be sufficient?

docs/snippets/glossary.md
```diff
@@ -0,0 +1,13 @@
+*[SED-ML]: Simulation Experiment Description Markup Language
```
jonrkarr, line 1: I think it would be helpful to sort this in alphabetical order

docs/index.md
```diff
@@ -198,3 +198,4 @@ The BioSimulators platform supports a broad range of modeling frameworks, model
 </a>
 </div>
 </div>
+--8<-- "glossary.md"
```
jonrkarr, line 201: Rather than embedding into every page, the index could be added to the menu via `.pages`

jonrkarr (changes requested):

docs/index.md
bilalshaikh42: > 
The glossary isn't actually displayed but provides tooltips over the words defined across the page. 
This page has a example of this extension https://squidfunk.github.io/mkdocs-material/reference/abbreviations/#configuration

PR #4178, opened 2022-02-07
fix(dispatch-service): bind the /tmp directory to /local on HPC

### LINKED ISSUE #4135: Bind /tmp to /local in sbatch script
Add the following to the singularity run command:
```
 --bind /local:/tmp
```
This should resolve the issue seen in #4127

bilalshaikh42:
the /tmp directory on the HPC can sometimes run out of space causing errors for some simulators.
This commit binds the /tmp directory on the singularity container to /local which is configured on
the HPC to work as a fast tmp directory.

closes #4135


jonrkarr (approved):
Looks good to me. Hopefully this reduces the frequency that jobs don't have temp space to work with.

PR #4181, opened 2022-02-07
fix: fix height for consent screen on mobile

### LINKED ISSUE #4143: Reduce the height on the cookie notification for mobile screens
Noticed by @moraru.

The height of the cookie notification needs to be constrained, such as `maxHeight: min(900px, calc(100vh - 2rem))` so that it can't be larger than the size of the screen.

This needs to be added at https://github.com/biosimulations/biosimulations/blob/0c1c222ee63a0e36857b93085f027c7f12fd41f7/libs/analytics/angular-analytics/src/lib/consent.service.ts#L51

bilalshaikh42:
the consent popup now has a max-height that depends on the screen height

closes #4143

### DIFF
--- a/libs/analytics/angular-analytics/src/lib/consent.service.ts
+++ b/libs/analytics/angular-analytics/src/lib/consent.service.ts
@@ -49,6 +49,7 @@ export class ConsentService {
       this.dialog.open(CookieConsentComponent, {
         hasBackdrop: true,
         maxWidth: 'min(900px, calc(100vw - 1.5rem))',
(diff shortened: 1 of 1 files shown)


jonrkarr (approved):
Haven't tested this, but I think this will force strolling.

bilalshaikh42:
> Haven't tested this, but I think this will force strolling.

I tested in chrome using the preset phone options as well as reposonsive. It does add scrolling if the screen gets too small

PR #4221, opened 2022-02-12
perf(api,dispatch-service): move downloading COMBINE archive and uploading to S3 to dispatch service

### LINKED ISSUE #4094: Refactor API to return response on simulation run creation BEFORE archive downloaded from a URL and saved to S3 storage
The present design imposes a bottleneck -- a large archive sent via a URL with limited bandwidth could take longer than the timeout limit to download and save to Google storage. I previously refactored this to the dispatch queue. This should be refactored to avoid this bottleneck.

Will likely be needed for large archives for VCell published models database.

jonrkarr:
closes #4094

### DIFF
--- a/apps/dispatch-service/src/app/submission/dispatch.processor.ts
+++ b/apps/dispatch-service/src/app/submission/dispatch.processor.ts
@@ -1,8 +1,8 @@
 import { SimulationRunStatus } from '@biosimulations/datamodel/common';
 import {
-  DispatchJobData,
   JobQueue,
   MonitorJobData,
+  SubmitHPCSimulationRunJobData,
 } from '@biosimulations/messages/messages';
 
 import { BiosimulationsException } from '@biosimulations/shared/exceptions';
@@ -23,7 +23,9 @@ export class DispatchProcessor {
   ) {}
 
   @Process({ concurrency: 10 })
-  private async handleSubmission(job: Job<DispatchJobData>): Promise<void> {
+  private async handleSubmission(
+    job: Job<SubmitHPCSimulationRunJobData>,
+  ): Promise<void> {
     const data = job.data;
 
     this.logger.debug(`Starting job for simulation run '${data.runId}' ...`);
@@ -41,7 +43,6 @@ export class DispatchProcessor {
       data.maxTime,
       data.envVars,
       data.purpose,
-      data.fileName,
     );
 
     if (response.stderr != '' || response.stdout === null) {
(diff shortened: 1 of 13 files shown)


apps/dispatch-service/src/app/submission/dispatch.processor.ts
```diff
@@ -91,4 +186,39 @@ export class DispatchProcessor {
 
     this.monitorQueue.add('monitor', monitorData, monitorOptions);
   }
+
+  private getRetryBackoff(): <T>(source: Observable<T>) => Observable<T> {
```
bilalshaikh42: I don't think we should add retrybackoff here. Ideally the uploading of the archive would be done as a separate workflow step. If this step gets into a retry loop for a while, it will cause all the job submissions behind it to also get stuck.

apps/dispatch-service/src/app/submission/dispatch.processor.ts
```diff
@@ -28,6 +42,87 @@ export class DispatchProcessor {
 
     this.logger.debug(`Starting job for simulation run '${data.runId}' ...`);
 
+    // resolve COMBINE archive and save to S3
```
bilalshaikh42: Throwing these errors here is not a good idea unless they are being caught by the  The API has a filter to catch these and return the proper https response. But the dispatch service is not returning anything, so these would end up as un handled promise rejections that cause the container to exit. This is the same issue that is causing the error with ssh.

apps/dispatch-service/src/app/submission/dispatch.processor.ts
```diff
@@ -57,8 +152,8 @@ export class DispatchProcessor {
       }
 
       throw new BiosimulationsException(
-        500,
```
bilalshaikh42, line 60: This is the error I was referring to that is probably the cause of the ssh issue. I haven't gotten to it yet, but this also needs to be replaced by the queue error handling logic

bilalshaikh42 (changes requested):
I think this should be architecture some thing like this: 

- Dispatch queue. Checks if url provided. If so, add a job to the upload queue. Else add a job to the HPC queue. Basic validation of incoming message. 

-  Upload queue. Uploads archive to s3, patches simulation with url, then adds job to HPC queue. 

- HPC queue. Submits the sbatch script 

This keeps all the steps atomic and idempotent allowing the error handling to be managed entirely by the queue system.

We'd then attach an error event listener to the HPC and upload queues that would mark the simulation run  as failed if either workflow step fails.

jonrkarr:
I moved the archive downloading and saving to S3 to a separate queue. I removed retrying because that will now be handled by Bull.

This is architecture:
- API
  - Uploaded archive: saves to S3 and submits to dispatch queue
  - URL for archive: submits to resolveCombineArchive queue
- resolveCombineArchive queue 
  1. Downloads archive
  2. Saves archive to S3
  3. Submits to dispatch queue
- dispatch queue (same as before)
  1. Submits job to HPC 
  2. Submits to monitoring queue

I'm not sure how you intend to handle marking jobs as failed. I mimicked how the dispatch processor handles this:
- Save failed status to database when maximum number of retries is reached
- Throw exception

bilalshaikh42 (approved):

PR #4280, opened 2022-02-19
fix(api): handled older simulation specifications that lack styles

### LINKED ISSUE #4279: 🐛: Existing runs broken by `SedStyle` changes
### Contact Details

_No response_

### Location

runBiosimulations, BioSimulations

### What happened?

Loading a page for a project causes an error message to display. 

### What made it happen?

Load https://biosimulations.org/projects/NFAT-translocation-Tomida-EMBO-J-2003

### Version

Production (Deployed on a .org domain), Development (Deployed on a .dev domain)

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
TypeError: Cannot read properties of undefined (reading 'forEach')
    at Af (sed-document.ts:77:27)
    at view.service.ts:395:32
    at Array.forEach (<anonymous>)
    at view.service.ts:393:34
    at map.js:7:37
    at c._next (OperatorSubscriber.js:13:21)
    at c.next (Subscriber.js:31:18)
    at Subject.js:31:30
    at c (errorContext.js:19:9)
    at c.next (Subject.js:26:9)
handleError @ error-handler.ts:36
...
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jonrkarr:
Closes #4279


bilalshaikh42 (approved):
Would it be more robust to add the same change to all the arrays? Have not run into any issues like that before, but seems to be a quick way to reduce the chance of other errors

jonrkarr:
I added the same for parallelism, although its unnecessary.

This happened because the data model changed, but the documents in the database haven't been migrated to the new data model. Because much of our data model is driven by the SED-ML and COMBINE archive formats, the data model is pretty stable except for upgrading to match new versions of those formats (which isn't happening with styles; styles are the first piece of SED-ML L1V4 I adding because BioModels needs it).

PR #4295, opened 2022-02-22
feat(dispatch): apply a new design to the dispatch component

jmdetloff:
Apply an updated design to the dispatch component, which is based more closely on material design.
Additionally, factor much of the code in dispatch.component.ts into functions.

Applies a new design to the dispatch component.

This PR does fix one bug I discovered while testing: In the current implementation the maxFileSizeValidator fails to correctly apply an error when a file that is too large is selected. It is checking control.value.size, which is undefined, instead of control.value.files[0].size. Additionally, changeProject uploads the file regardless of whether this error is applied in the current implementation. Both of these errors are fixed with the implementation in this commit.

I did include large style changes in this commit. Apologies, I know that makes for a harder review. I can attempt to break them out into separate commits later this week if you'd prefer. Adding comments and breaking code into functions is a part of my process to ensure I understand what I'm reading, and in this case I've left those changes in since they should improve general readability and testability of this component as well.

- Many of the large functions within dispatch.component.ts have been broken into smaller functions.
- Many large expressions within dispatch.component.ts have been broken down into multiple smaller expressions.
- Class variables that are redundant or can be local variables have been removed from dispatch.component.ts.
- I've added comments in a few places I felt they could help.
- I've updated linebreaks and indentation style to match Google's style guide. I hope this isn't an unwelcome change. I saw a couple format prs go by that made me think the more aggressive line breaking was being removed going forwards. I can reset this if desired.

I've run existing dispatch tests, but have not written new ones for the dispatch component. A lot of the dispatch component is a little too UI focused for effective unit testing, but in a future commit it would be great to break the logic for loading data and updating model into a separate controller class that could be more easily tested. I could update this PR with another commit to implement new tests later this week if we want to get those ready before merging.

For now I've manually exercised that the controls update as intended, that errors display when and only when inputs are incorrect, that controls populate as they should when opening the route with query params, that controls enable and disable options as model, algorithm, and policy selections are made, that controls populate as they should when a sed doc spec for a selected project loads, and that form submission completes successfully.

Here is a video displaying what the new UI looks like:

https://user-images.githubusercontent.com/1418389/155122999-cf90b41f-bcc3-42a0-949d-692ad4d8fb61.mov


jmdetloff:
We discussed applying this design to the other forms within dispatch, and of course to the new edit simulation form. I haven't had time to finish these changes yet but hope to later this week. I can either update this PR with additional commits or create new PR's for these changes as preferred.

bilalshaikh42:
> I did include large style changes in this commit. . Apologies, I know that makes for a harder review. I can attempt to break them out into separate commits later this week if you'd prefer.

No worries, this looks great! I think the code is much more readable this way

> * Many of the large functions within dispatch.component.ts have been broken into smaller functions.
> * Many large expressions within dispatch.component.ts have been broken down into multiple smaller expressions.
> * Class variables that are redundant or can be local variables have been removed from dispatch.component.ts.
> * I've added comments in a few places I felt they could help.

Thank you so much for this. This is definitely an area we have not focused on as much as I would like since we have been rapidly prototyping. This is really helpful


> * I've updated linebreaks and indentation style to match Google's style guide. I hope this isn't an unwelcome change. I saw a couple format prs go by that made me think the more aggressive line breaking was being removed going forwards. I can reset this if desired.

So we use prettier for formatting, so this might actually end up getting autoformatted out when committing. Generally, running 'nx format' will format all the files to the settings in the `.prettier.rc`. If you have any preferences to change the prettier settings, feel free to make a PR. Prettier ensures that each dev does not need to think about formatting much. Regarding the line breaking, that's on me. We have a warning for lines that are too long, so in an attempt to make those better I accidentally used multiline strings which added a bunch of whitespace in the middle of the line. The changes you referenced were probably me undoing that and switching to using string concatenation instead.

I really like the new clean look of the UI. Do you think the user experience is better without the icons? That is the only thing that caught my eye that I thought was good to have in the older design. Or was it making the UI too cluttered/busy? I don't know much about the actual principles of UI other than personal opinions

jonrkarr:
The new style looks nice! We're not working on the UI very much, so I think its fine to group changes to all of the forms (at least the existing ones) into 1 PR so they are all deployed with the same style together. We can merge those changes in when you're ready.

Improving the UI will be a big improvement for users. After the content of the database of modeling projects, a UI for modifying simulation projects and re-running them is the main requests we've gotten. As Bilal alluded to, we've focused our attention on building the abstractions, building backend services to run simulations, and curating content (simulation methods, models, data visualizations). We haven't had as much bandwidth as we'd like for the UI.

jmdetloff:
Thanks!

I'll take a look at the prettier settings. You're right running format does re-format a lot of my changes. Does the team currently have strong preferences about formatting? I tend to lean towards one-lining expressions whenever they can fit in the line limit, as long as the information content isn't too out of hand. A bit of flexibility can be nice with formatting I find. Making simple areas dense and complex areas more sparse can help a reader understand what is boilerplate they can skim and what is meaty code to read closely, but I understand consistency is valuable as well.

As for the icons, I did find the previous UI a little too dense, and that removing the icons was an improvement to that. However I'm no designer, I'm just operating from intuition and personal preference, so take it with a grain of salt! One bit of reasoning I'll add here is that a lot of these icons either aren't instantly recognizable, aren't closely related to the function of the form section, or are redundant.

For example the algorithms, models, and policy fields are all right next to each other and all have the same paper icon. So this is using a fair amount of space, is redundant, and it's not an icon that I personally would immediately recognize or that would help me understand the purpose of the field. The simulation tool section has a gear icon, which is related to a tool, but it's not an icon I recognize or that would help me understand this field either more quickly or more correctly than by just reading the description alone. In cases like this my preference is to lean towards a more minimal design, where call to actions and text content are front and center, and icons that don't have an immediate value add are trimmed. 

But I could certainly create a version with the icons re-added so they could be compared if you'd like, or simply add them back in!

jonrkarr:
I don't have strong preferences about format. Bilal set up the automatic TypeScript formatting. The Python formatting follows PEP8. Line length is one of the style rules I've overrode for Python because I find the default too short.

I agree that the icons are closely associated with the domain content. We've largely stuck with built in icons from FontAwesome. Ideally, we'd create a more informative icon set. Removing undifferentiated, uninformative icons makes sense.

bilalshaikh42:
> I'll take a look at the prettier settings. You're right running format does re-format a lot of my changes. Does the team currently have strong preferences about formatting? I tend to lean towards one-lining expressions whenever they can fit in the line limit, as long as the information content isn't too out of hand. A bit of flexibility can be nice with the formatting I find. Making simple areas dense and complex areas more sparse can help a reader understand what is boilerplate they can skim and what is meaty code to read closely, but I understand consistency is valuable as well.

No preference really as long as it's consistent.  I think? the settings for the prettier are just the prettier defaults, which are somewhat widely used, but if you have any preferences that can be added to the prettier settings file, just run `nx format` to apply to all the files.

Regarding that using density would separating out boilerplate blocks with extra line breaks and comments work?

jmdetloff:
Just pinging here to let you know I haven't forgotten about cleaning this PR up and getting it ready to merge! My team got caught up in a re-org at Facebook. The dust hasn't settled yet and my time/energy has been a casualty. I should be able to carve out a slice of time early this week. If this is blocking any tasks that you'd like to start asap let me know and I'll make this a priority!

jonrkarr:
Thanks for the update. This isn't blocking anything. We're still trying to wrap up integrating more models (content for the database).

jmdetloff:
I've added a few more commits to this PR tonight, which cover my work so far towards applying the new design to the Create Project form. I'm afraid to say I also went a little beyond scope here again and made an additional UI tweak. I noticed there was a lot of logic in create-project.component.ts around updating and clearing field values when other field values are modified by the user such that they are no longer compatible.

This is unideal because it can be disorienting to a user to see things changing in their form that they don't expect, the logic is complicated and there were broken edge cases, and it allows for the potential for a user to create a project with different values than they thought they entered. And worse, when the introspection request completes it auto-fills several forms without warning, potentially overwriting values the user may have entered without them knowing.

To solve this, I thought this form might be better represented as several sequential steps that flow from one to the next. If one field depends on another, then it comes in a later step, so that by the time the user sees any given field and has a chance to edit it there is no longer any possibility of it becoming incompatible with its dependencies and needing to be cleared or reset. Additionally, a spinner is inserted into the flow while introspection takes place, so fields are prefilled before the user sees them.

The end result is a decrease in the amount of syncing logic that has to be maintained in code, and IMO a better user experience, but of course that is subjective. This can all be put back into a single page form with a little work if thats the design you'd prefer!

Other than this, the design changes are similar to the last round: increasing margins and padding, increasing font size, removing header backgrounds and some borders and icons, generally giving the form more space.

Here is a video of me doing a quick test run through the form:

https://user-images.githubusercontent.com/1418389/159645280-6f2d3f5e-663f-4d8b-a0f9-f57e587d6cf8.mov

Similarly to last round, I've also done some code cleanup as I've gone through the relevant files. This amounts to a lot of breaking things into separate functions, creating a few shared lips for decreasing boilerplate, and breaking out a lot of the logic for translating the data model from combine classes to dispatch classes and back into a utility class of static functions.

This will hopefully improve readability and definitely improves testability. I plan on doing more manual testing with this commit in the near future, as well as documenting the new class/functions, and writing tests. The logic for paging the forms could also be broken into a shared form paging component lib if it needs to be used elsewhere. So this isn't quite ready to land yet and if you pull for local testing you may find broken edge cases, but I figured I'd get some feedback on the design changes before doing any more polish.

Does the multi-step form design seem like a good fit? Any other changes you'd like to see added or reverted?

jmdetloff:
Additional important note: I added a commit here to update the prettier settings. The change sets the new line guide to 120 characters per line, and makes it so brackets stay on the same line when prettier inserts newlines. I found these settings to provide a more comfortable experience, but let me know if you disagree and I'll set them back.

jmdetloff:
@jonrkarr Thoughts on the multi-step form approach vs maintaining the single page form? I'll have some time for tests and polish this weekend, but want to make sure that approach is ok before continuing.

jonrkarr:
Thanks for these suggestions! Breaking the steps down is nice. I agree this is probably much easier to understand. The videos are really helpful.

(a) Could you change the order of the steps
- (No change): Choose model format / model
- (No change): Chose modeling framework / type of simulation / algorithm
- Review/modify namespaces
- Review/modify model changes
- Review/modify which variables should be recorded
- Review/modify algorithm parameters

(b) Could you display the algorithm parameters and compatible simulation tools in the same pane? Because this is related, I think this is helpful to display together.

(c) The very last buttons ("Simulate" and "Download") are much larger than the next/back buttons. Should these be stylistically congruent?

We should probably merge this soon to ensure there's no painful merging (not that we intend to make any changes related to this soon).

jonrkarr (approved):

jonrkarr:
@jmdetloff please let me know when you'd like this merged.

jonrkarr:
In case its not obvious, the CodeFactor and SonarCloud failures can be ignored. We're not following those strictly.

bilalshaikh42:
The multistep form is much better! I had been hoping to make these workflows multistep forms as well, definitely agree its a better user experience. 
The sonarcloud checks did pick up two potential bugs that should be fixed though. Let me know if they are false positives and we ca then ignore them

bilalshaikh42 (changes requested):
Looks good pending the two potential bugs
The security hotspot is a false positive since we use 'http' for identifiers, not 'https'

jmdetloff:
Woops, looks like those two bugs are both referring to a typo that slipped in, I'll get rid of that and resubmit. I also have a couple commits with tests incoming. After those arrive this should be good to land!

Can definitely move around the order of the forms and combine algorithm parameters and simulation tools. 👍

As for the final buttons, I went back and forth a bit on the styling. On one hand they look a little odd with the extra space they create in the layout, on the other hand they are the main call to action for the form so emphasizing them made sense to me. I also tried them center and left aligned, and stretching to span the width of the pane. I'll try another version with them smaller to match the next/back buttons and post a screenshot.

jmdetloff:
Another quick question: The run flow could be factored into a multi-step form as well if that's the UI you'd prefer. Should I make that change also? If I recall correctly the fields aren't quite as interdependent on that one, but there are some selectors whose options update or clear in response to other fields being edited, so multi-step could have value there as well.

bilalshaikh42:
> Another quick question: The run flow could be factored into a multi-step form as well if that's the UI you'd prefer. Should I make that change also? If I recall correctly the fields aren't quite as interdependent on that one, but there are some selectors whose options update or clear in response to other fields being edited, so multi-step could have value there as well.

Up to you.
When a file/url is added there is some introspection that disables the simulation tools that are not compatible with that model. This could cause edge cases like the one you noticed earlier when the model is changed, so if you think it would be better to make it modular go for it. 

On the other hand, as someone who often uses the form to test the backend,  having to do multiple clicks to step through might also be annoying.  

I'll defer to you on this, the changes so far have been great, so whatever you think works best. 

If it isn't too much trouble, #4390 could be included as well. When I am submitting multiple runs to test the backend typing out names gets annoying.  Making the name optional entirely would be possible, but this is incorporated into the UI/API/database models, so would need changes throughout. 
 
I thought a better method might be to have the form generate a random name if the user does not provide one, kind of how docker creates an "adjective_scientist" name for containers. Unsurprisingly, someone has made an npm package to do this. https://www.npmjs.com/package/randimals

jmdetloff:
Thank's for your patience with this PR! I've been team transitioning over to PyTorch at Meta, so ramp up study has eaten a lot of my time the past few weeks.

I've pushed a new commit to this PR this morning in which I make some changes/additions:
- The changes @jonrkarr requested have been made: The steps are reordered, simulation tools and algorithm parameters exist in the same form step, and the style of the simulate and download buttons have been changed to match that of the next button.
- The logic related to paging through the form steps has been factored out into a new component, MultiStepFormComponent. This will allow for easy re-use if needed in the future, clearer code in the create-project component, and isolated tests of the form functionality.
- A good deal of static logic related to creating a CombineArchive from form data, determining which project options are compatible with each other, and introspecting a SedDocument have been broken out into functions in the modules archive-creation, compatibility, and introspection respectively, within the simulation-project-utils-service project. This logic is all state free, so moving it into functions will make for easier reading and testing, and allow the components that use them to focus more on presentation logic.

There's a lot of code moving around here, but the cumulative effect of factoring out the form paging logic into MultiStepFormComponent, creating individual sub-components for each step of the form, and moving static logic into new modules is that the create-project.component.ts file has shrunk down from an intimidating ~1900 lines of code to a much more manageable ~180 lines of code. The newly created components and modules are of a similar scale.

I've written some tests, and stubbed out some more, but this PR has ultimately touched quite a bit of code. So, there are many more tests that could and should be written. I'll keep adding them as long as this PR is open, but IMO this should be solid for landing in its current state. I've run it through the paces manually, and will attach a video. I've been testing primarily by using the model at http://bigg.ucsd.edu/static/models/e_coli_core.xml, playing around with the different form settings, and then running the simulation or downloading. If there are other important flows I should make sure to exercise let me know and I can run through them and attach videos as well. 

What is still not included in this PR, but can be added in follow up PR's:
- The run flow has not been changed to a multi-step form. @bilalshaikh42's comment about this adding friction to a flow that is run through frequently is a good point. In this case the fullscreen form might strike the better balance between clarity and efficiency. This can always be changed down the line if needed.
- #4390 I'll dig into this next, and either add it to this PR, or spin up a new PR if this one has already landed.
- Complete test coverage - I've written some tests for the new MultiStepFormComponent, and a handful more for the create project form steps, but there are plenty more tests to write to achieve full coverage. The good news is that testing should be much easier going forwards with the more modular code.

jmdetloff:
https://user-images.githubusercontent.com/1418389/167296702-d98ad6d3-6eea-405c-8efb-7a87a9ad2e12.mov

bilalshaikh42 (approved):

bilalshaikh42:
> What is still not included in this PR, but can be added in follow up PR's:
> 
> * The run flow has not been changed to a multi-step form. @bilalshaikh42's comment about this adding friction to a flow that is run through frequently is a good point. In this case the fullscreen form might strike the better balance between clarity and efficiency. This can always be changed down the line if needed.
> * [Allow simulations submission without specifying a name  #4390](https://github.com/biosimulations/biosimulations/issues/4390) I'll dig into this next, and either add it to this PR, or spin up a new PR if this one has already landed.
> * Complete test coverage - I've written some tests for the new MultiStepFormComponent, and a handful more for the create project form steps, but there are plenty more tests to write to achieve full coverage. The good news is that testing should be much easier going forwards with the more modular code.

@jmdetloff 
I have merged this in to get some conflicts with #4451 resolved, so feel free to open additional prs for these! Thanks again.

jonrkarr:
The improved form looks great!

The most helpful contribution would be the UI we discussed for editing simulation projects.

The ability to submit projects without a name isn't a priority. Plus, we could easily address this by automatically generating a name such as `{simulation-tool-id} {simulation-tool-version} run {time-stamp}`.

Complete test coverage would be great, but this is already better than we have for nearly all of our UI components. We have extensive testing for the underlying simulation logic, but limited testing above that.

Issue #4319, opened 2022-02-26
45 of 74 ModelDB runs did not succeed for multiple reasons


jonrkarr:
The first number in each bullet is the ModelDB id of the project. The corresponding COMBINE archives in https://github.com/biosimulations/biosimulations-modeldb/tree/dev/biosimulations_modeldb/final/projects.

- [x] Thumbnails in sub directories could not be saved -- See #4318 
  - 31 projects
- [x] Failed to extract file -- see #4329 
  - Unpacked contents of archive not uploaded to S3
  - This error is reproducible.
  - Examples 
    - 229640: `621958e10df8c75a0ac00ca7`
    - 152028: `621958d12f0784cadefe366b`
    - 206365: `621958da0df8c75a0ac00bf4`
  - These are the archives with the largest number of files (258, 331, and 3705). 
  - The biggest successful archive have 109 files
  - The extraction appears to fail when archives have many files
- [ ] Stuck in post-processing
  - May be due to failures with extraction
  - Still running a day later -- this reveals that there's also issue with error handling
  - Example
    - 228337: `621958df2f0784cadefe3730`
- [x] hsload failed
  - This does reveal an issue with hsload -- reported as HDFGroup/h5pyd#116
  - However, the real problems for us aren't hsload, but failures to correctly handle XPP
  - Examples  
    - [x] 83570: `621958bc2345ec060040d0b8` -- simulation failed because not executed from correct working directory
    - [x] 185334: `621958d50df8c75a0ac00bb4` -- invalid model file
    - [x] 227577: `621958dd0df8c75a0ac00c39`-- simulation results because not executed from correct working directory
- [ ] Failed to publish -- does this need to be retried?
  - 58195: `621958b50df8c75a0ac00a9c`
  - 145800: `621958cd2f0784cadefe3641`
  - 151677: `621958d02f0784cadefe3664`
- [ ] Failed to updated published run -- does this need to be retried?
  - 35358: `621958b22f0784cadefe3595`
  - 45513: `621958b30df8c75a0ac00a94`
  - 57910: `621958b32f0784cadefe359c`
  - 58172: `621958b40df8c75a0ac00a97` 

## Example hsload failure
```
Traceback (most recent call last):
  File "h5py/h5o.pyx", line 302, in h5py.h5o.cb_obj_simple
  File "/home/FCAM/crbmapi/.local/lib/python3.6/site-packages/h5py/_hl/group.py", line 591, in proxy
[...]

Issue #4283, opened 2022-02-19
Improve performance of COMBINE API 


jonrkarr:
- [x] Speed up JSON encoding. Done in #4288.
- [x] Extract specifications of manifest from manifest file (in S3) rather than COMBINE archive file: ~0.3 s. 
  - [x] Enable COMBINE-API to extract manifest specifications from individual manifest XML files in #4285 
  - [x] Change manifest queue to depend on the extraction queue
  - [x] Change dispatch service to send the S3 URL for the manifest XML file rather than the URL for its parent COMBINE archive (required code is commented out and just needs to be uncommented)
- [ ] Speed up extraction of metadata: ~0.5s for iAF1260
  - Read from individual S3 files rather than COMBINE archive; requires passing manifest or path to manifest
- [ ] Speed up extraction of SED-ML specifications: 2-5s (mostly content download)
  - Read from individual S3 files rather than COMBINE archive; requires passing manifest or path to manifest. At least for modest size archives, downloading the archive appears to be a small part of the overall duraton.

## Performance data
I ran SED-ML extraction for all 92 BiGG models. The average (using https://combine.api.biosimulations.dev) was 3.3 s, ranging from 0.2-8.0 s. This includes the time to transfer the result to my machine. It should be less from the COMBINE API to the dispatch service. For localhost to localhost I get 0.1-1.4 s (avg 0.9).

The core SED-ML extraction is only around 0.2 s on my machine (max ~0.4 s). Half of that is libSED-ML.

jonrkarr:
@bilalshaikh42 I looked into the performance of the COMBINE API SED-ML extraction. There doesn't seem to be anything obvious to improve with the code. One other thought is to increase the CPU assigned to the COMBINE API (e.g., the deployment is much slower than my local machine).

https://github.com/biosimulations/deployment/blob/aa2ba78a1484e06728c97b40654ab06cf8968b77/base/apps/combine-api.yaml#L24

bilalshaikh42:
I am not seeing the cpu usage for the combine-api come anywhere near  the limit, so that won't make much difference. The only other reason that I can think of is that it has something to do with the gunicorn wrapper

jonrkarr:
My recollection is that you mentioned that the combine api is going down. It seems that the CPU time and load are fine.

I could take another look at the gunicorn settings. However, I'll be grasping at straws. The settings aren't particularly clear.

Is there any log information that is illuminating?

PR #4329, opened 2022-02-26
refactor(dispatch-service): sped up extraction of files from COMBINE archives, especially for large ones

### LINKED ISSUE #4112: Investigate how to speed up extraction
The extraction of the combine archive is the bottleneck in processing. There are a few ideas on how to fix this. Getting some empirical data would be helpful 

1. Try to start the extraction before the processing step, but still maintain the dependency for the other steps
2. Move the extraction to the HPC
3. move extraction and additional processing steps (one that don't need the simulation run to complete, essentially all but the logs, output) to a separate workflow that starts when the job is dispatched.

### LINKED ISSUE #4321: `ExtractionService` in `apps/dispatch-service/src/extraction/extraction.service.ts` is not used
In trying to debug #4319, I noticed `ExtractionService` is not used. This should be used or removed.

### LINKED ISSUE #4326: `extractZipObject` fails on COMBINE archives with many member files
As noted in #4319, `extractZipObject` appears to fail with archives that have a large number of members likely 128 or 256. This is irrespective of the size in bytes of those files.

### LINKED ISSUE #4327: Change `extractZipArchive` to optimistically save files to S3
`extractZipArchive` currently terminates on the first failure rather than uploading as many files as it can and then terminating with an error. The issue is due to using `Promise.all` without additional error handling.

### LINKED ISSUE #4328: `unzipper.Open` is very slow for archives with many files
The unzipper package seems to be quite slow for some archives, such as modeldb 152028 (258 files). Using an uncompressed Zip file doesn't seem to help much.

An easy solution is so migrate S3 uploading to the HPC.

jonrkarr:
Dramatically speeds up extraction/saving to S3 of larger archives

Closes #4328
Closes #4112
Closes #4327 
Closes #4326
Closes #4321
Addresses part of #4283
Addresses part of #4319


jonrkarr:
Merged manually due to conflicts

PR #4365, opened 2022-03-19
fix(dispatch): ensure params can be fetched for models provided by url

jmdetloff:
Currently a bug prevents the get-parameters-variables-for-simulation from being hit when uploading a
model via url through the create project form. This commit fixes this issue.

N/A

When location type is set to url, then the file will always be missing, and this conditional will always resolve to false. The result is that when using the create project flow, and uploading a model via a url, the algorithm parameters, model changes, and model namespaces will not be preloaded. This could result in users uploading models with incorrect settings or failing to complete the form at all.

N/A

- I used the production create project form and confirmed that no network call to get-parameters-and-variables-for-simulation occurs when using a model url. 
- After making this change, I tried the same process again, and confirmed that the network call is sent and that a SedDoc is returned.
- I followed the process again with the same file but uploaded locally and confirmed that the network call is still sent and returns successfully in this case as well. 

**Additional information**
I have a commit to redesign this form and do some code clean up coming, and discovered this bug while testing it. It seemed impactful enough that I should break it into a pr of its own.

### DIFF
--- a/libs/simulation-project-utils/ui/src/lib/create-project/create-project/create-project.component.ts
+++ b/libs/simulation-project-utils/ui/src/lib/create-project/create-project/create-project.component.ts
@@ -634,20 +634,17 @@ export class CreateProjectComponent implements OnInit, OnDestroy {
       this.modelParametersAndVariablesSubscription = undefined;
     }
 
-    const modelLocationTypeControl = this.formGroup.controls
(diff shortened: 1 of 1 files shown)


bilalshaikh42 (approved):

bilalshaikh42:
Thanks!

PR #4466, opened 2022-05-14
fix(dispatch): fix data generator id set during archive creation

jmdetloff:
Ensure the id set on the data generator created during frontend archive creation matches
'data_generator' + the id of the data set

I noticed this small error when diff'ing an archive downloaded from the old project creation flow with one from the reworked flow. DataGenerators originally would be named like: 'data_generator_flux_reaction_R_ACALD' and after my changes they were named like: 'data_generator_variable_flux_reaction_R_ACALD'. After this fix they will once again be named without the extraneous '_variable'.

Im not aware if the _variable token in the id actually causes any issues, but I figured best to revert to the original behavior just in case.


bilalshaikh42:
Thanks!

PR #4472, opened 2022-05-27
feat(platform): cleaned up summary stats view

jonrkarr:
- Moved the biology section to the top, formats/tools/algorithms to the middle, and sources/licenses to the bottom
- Used "&" and "and" consistently in the section headings
- Added alternating background to sections
- Corrected width of sections
- Used consistent casing and pluralization in the section headings
- Made the section subheadings have an equal number of lines so that the plots are positioned consistently
- Cleaned up plot spacing when there are only 2 occupied columns
- Removed the titles of the plots; they're redundant with the section headings
- Moved the modeling formats plot to the second section with methods and tools
- Used our color scheme
- Corrected capitalization of options attribute

@bilalshaikh42 todo
- [x] The versions of dependencies appears to be corrupted. I needed to install with the `--force` option. This needs to be fixed. Note, this PR adds a new requirement (`'@thi.ng/color`) and removes `google-palette`
- [ ] Put the pie chart item labels in the pie. If the labels are long, use a bar chart
- [ ] Use bar charts rather than pie charts when there's >= 6 items


jonrkarr:
@bilalshaikh42 thanks for correcting the tests are merging this in.

PR #4581, opened 2022-09-25
refactor(dispatch): consolidate forms modules into a single directory

jmdetloff:
This is a small organizational diff I've been working on top of locally for a bit, which can land on it's own. It consolidates form host, validators, and multi step form into a single shared ui directory. It also renames MultiStepFormComponent to PagingFormComponent, as I have follow up diff's that factor some form logic into a MultiStepForm component that is independent of the paging behavior (Useful, for example, for multi step forms where all the steps are on the page at the same time, but for which we still want to update the options on each step when the user selects an option on any single step.)


jmdetloff:
Looks like the shared-ui lib has a lint error. Just checked and this error does seem to exist on top of the dev branch

bilalshaikh42:
Looks great, thank you!

PR #4582, opened 2022-09-26
[WIP] Add the modify archive flow to the dispatch form

jmdetloff:
This isn't ready to merge yet - however I've had this issue in progress for a long time so I'm pushing up a few commits and screenshots just to provide an update on the progress.

I've factored the dispatch component to use an abstracted form component, just like the create project form. This allows for a lot of form logic re-use, and also breaks the form steps into their own smaller components. Then, for cases when the project (and optionally simulation tool) are preselected in the url params - which is the case when users tap the re-run button for simulation runs - the form steps to select an archive, specify capabilities, and select a simulation tool are replaced with a single form card that displays the current preselected options along with two buttons. One resets the form to its un-preselected version with all the form cards visible and empty. The other launches into the modify flow, and is only visible if the selected archive has a single model and a single simulation task. Here are a couple screenshots showing the new two modes of the dispatch form:

The form state seen when accessing "/runs/new" with the archive and sim tool defined in the query params (when clicking rerun on a simulation run)
RerunState

The default state seen when accessing "/runs/new"
DefaultState

I expect to stack one more commit on this PR, which will be the implementation of the modify route and component, and is still WIP. It makes heavy re-use of the factored out form steps from the create project form, but only displays a subset of them, and utilizes the new srcHandlersCombineModifyHandler api in the combine api angular client library on submission.

Let me know if there are changes to the approach or UI that I should consider here, and feel free to close out this PR so it doesn't create clutter, I just wanted to provide a brief update here. I'll hopefully have the modify component complete in the near future and will either stack that commit on this PR or create a new one.


jmdetloff:
I've stacked another commit on this PR this morning, which contains the UI for the form to modify and archive before rerunning. The modify form contains form steps for modifying the uniform time course parameters, the model parameters, and the algorithm parameters. The algorithm parameters form step is simplified compared to the form step in the create project form, because it isn't preceded by the form step for selecting a model format and framework, so the api's for fetching compatible simulators weren't immediately compatible. A follow up diff could fetch more information about the algorithm parameters from the api with the algorithm id.

This commit does not contain the code to actually modify the archive when download or simulate is tapped. This is the last step required, other than any UI changes you'd recommend, so I'll be adding another commit to accomplish that to this PR before it will be ready for merging or a close review - this is just an update.

Here's a quick video of the flow as of this commit:

https://user-images.githubusercontent.com/1418389/193455417-bc946442-d79f-448f-8c4c-6f90c1b4daf3.mov

jmdetloff:
I pushed on another set of changes this morning. The topmost commit now creates the modified CombineArchive from the original archive and the user input from the run with modifications form. However, it looks like the implementation of srcHandlersCombineGetFileHandler isn't quite finished - it doesn't add the specs to the post request, so the request 400's. 

I'll be able to complete that in the near future. In the meantime I've implemented everything in this PR and just comment out the final call into srcHandlersCombineGetFileHandler. Once that function is complete this PR should be good to land, after some thorough testing and writing some docs. And it should be in good enough shape at this point for a review of the broad strokes.

PR #4624, opened 2023-03-21
chore: build update dependencies and migrate apps

bilalshaikh42:


jcschaff:
@bilalshaikh42 I'm finishing up by combine-api branch, I'll have to merge without all tests passing unfortunately.  

**If you can merge this PR soon, I can start the Angular development from here, otherwise I'll cherry pick as needed.**

I'm starting a simpler, unified Angular app, pulling good things in as needed (without circular dependencies, with fewer modules, and more off the shelf solutions - e.g. more emphasis on filtering than progressively loading 500 models).

bilalshaikh42:
@jcschaff This is largely ready, but it completely breaks the styling of the app. This is because we were not disciplined and used a lot of private implementation details and classes to style things, which broke with the update. If you are planning on starting a new app, we can merge and fix the style as we go along on that one

jcschaff:
I'm starting to review your changes, mostly so that I can follow along and learn.  If we can remove the circular dependencies and get everything to pass the unit tests, then I can simplify the existing design in-place rather than start over.

With a weeks effort, can this repo be made to pass all tests?

bilalshaikh42:
Yup, should  be doable. I think the main issue with the tests is the circular dependencies that you identified. Could you point me to that?

bilalshaikh42:
> I'm starting to review your changes, mostly so that I can follow along and learn. If we can remove the circular dependencies and get everything to pass the unit tests, then I can simplify the existing design in-place rather than start over.

The first commit is the most involved one. That one I actually didn't do myself, but ran the `nx migrate` command. This updates all the packages that nx manages, as well as makes automated changes to account for breaking changes to the code. 
The [docs](https://nx.dev/core-features/automate-updating-dependencies) explain this process a little bit more

Unfortunately, this was not fully successful because of a couple of factors:
1)  we had been abusing the CSS classes for the angular material library which broke some of the sass files we have
2) We were using an experimental web API (the resize observer) through a pollyfill library but this is now built in to the latest framework
3) We updated typescript compiler, which is now catching errors that it was not able to before. Technically the errors were always there, but now that the compiler is  "smart" enough to notice them, it breaks the build
4) The api client libraries (combine-api-client and hsds-client) are built using code generation that targets the older version of Nestjs. These are minor changes to make manually (and i did in commit fe391fbfd8665210b96b720286949032002c5cfe) but long term we need to use an updated version of the code generator tool to regenerate these clients

Usually, migration is a lot smoother, but we hadn't run one in several months

bilalshaikh42:
Closing in favor of #4640

bilalshaikh42:
Closing in favor of #

PR #4665, opened 2023-04-14
feat: integrate runBiosimulations into main platform webapp

### LINKED ISSUE #4664: feature(platform): integrate run.biosimulations.org as biosimulations.org/run
remove external routing between main biosimulations.org site and the run.biosimulations.org.

jcschaff:
see #4664

local testin


bilalshaikh42:
Rather than removing the run app entirely, we can integrate the functionality into the main app without removing the existing run domain. I think this is important to maintain the site as presented in our publication. 

The run app is also far simpler than the platform app, which needs a lot of reworking of the UI as we discussed, so that might be a candidate for a rewrite from scratch rather incorporating run into it

jcschaff:
@bilalshaikh42 or @jonrkarr : in addition to code reviews, it would be great if you could take a quick look at the CSS problems - I tried to roll back to Angular 14 to avoid these - but ran into more dependency hell.  For example, the svg nav button rendering and layout is now wrong.

jcschaff:
SONAR cloud complaints about code duplication is intentional - copied code from runBioSimulations into Biosimulations, with the intent to freely modify the copy while retaining the original runBioSimulations.

bilalshaikh42 (approved):

bilalshaikh42:
The css is going to be tricky. Lots of the internal classes have changed. My css is quite bad but I'll take a look and try to get the more basic components working well. 

The more advanced components like the table view we probably can style as we refactor them.

jcschaff:
> The css is going to be tricky. Lots of the internal classes have changed. My css is quite bad but I'll take a look and try to get the more basic components working well.
> 
> The more advanced components like the table view we probably can style as we refactor them.

@bilalshaikh42 :  I agree, the basic styling of new components will be easier.  Any partial repairs you can make to the worst of the CSS problems would be appreciated (e.g. home page navigation buttons).

PR #4674, opened 2023-04-23
feat: added link from project table to project detail

jcschaff:
provide navigation from the project table to project detail view.

tested locally


jcschaff:
initially tried to fold in client-side filtering into this PR, but it won't play very nicely with the server-side search/pagination - and the existing implementation is not simple.  For another PR, we will implement a server-side filtering capability.

### A new strategy for filtering in next PR

a single integrated solution for server-side text-search/filtering/pagination:
Once upon 'startup':

1. Server - generate list of filter categories and known values across all existing projects (could be maintained as a mongo collection).
2. Client - retrieve list of filter categories and valueus from server and dynamically generate client side UI.

Upon each table data change event (change page, search term, filter choices):

1. Client - send selected filter options along with the request - and filtering is done server-side before pagination.
2. Client displays whatever data the server returns.

moraru (approved):

PR #4675, opened 2023-04-25
feat: server side filtering of project table

jcschaff:
restore categorical filtering of projects in the project browse table in the platform app.  I will be difficult to use the existing filtering widget which does client-side filtering, as the table is architected differently, and it would be difficult to reconcile the filtering with server-side paging.

the previous project browser allowed users to filter by category, but this was lost in the new scalable project table implementation. 

rollup of depend-bot angular tools upgrade included.

Will be tested locally and a server-side unit test will be introduced.


jcschaff:
the client-side filtering offered by the previous implementation is difficult to reuse, and won't play nicely with server-side pagination without significant changes.  It does provide some good inspiration.

We need a single integrated solution for server-side text-search/filtering/pagination:
* Once upon 'startup':
  * Server - generate list of filter categories and known values across all existing projects (could be maintained as a mongo collection).
  * Client - retrieve list of filter categories and valueus from server and dynamically generate client side UI.
* Upon each table data change event (change page, search term, filter choices):
  * Client - send selected filter options along with the request - and filtering is done server-side before pagination.
  * Client displays whatever data the server returns.

moraru (approved):

PR #4687, opened 2023-05-31
fix: ui adjustments and linking out to prod site

AlexPatrie:
_**What does this PR do?**_

Adds pagination to top of projects table, and more UI style overhauls.


jcschaff (approved):
verified Alex's UI changes and the links pointing back to the production site.

PR #4694, opened 2023-06-02
fix: adjust nav button behavior in app

AlexPatrie:
_**What does this PR fix?**_

- Fixes unfavorable interaction behavior of buttons by adjusting `timeout` to increase by a factor of 10 in `window.setTimout()`
- Fixes issue of clickable area of interaction on dropdown menu.


jcschaff (approved):
looks good, approved.

PR #4700, opened 2023-06-07
fix: updating projex app to include corrections

AlexPatrie:

* Exposing simulation-run files to the simulation profile page
* Exposing project-table search to more or ’s . For example, in searching for the full term “Lorenz-system-Garny”, the project table returns None if the search term is simply “Lorenz”.  As users begin to type “Lo”, that should be one of the search results that is returned. (Not as easy to explain, but is a very simple issue)
* Push down vis col in simulation profile page as per the new styling
* Working to sync paginator in project table to return the accurate amount of entries
* Tinkering with re-running simulations


jcschaff (approved):
will approve without a detailed look

jcschaff:
merge when ready

jcschaff:
okay, turned off parallel build test lint in CI and it passed the action - 20 minutes is not terrible if it works reliably

PR #4719, opened 2023-06-25
fix: project list pagination with filtering showed no projects

jcschaff:
### Problem
The Biosimulations projects page displayed a blank project list when a filter is used without a search term.
### Solution
fixed project api for filtering, page size and page index were transposed.

Tested manually, locally.  Proper unit tests are needed.


AlexPatrie (approved):
Looks good!

PR #4722, opened 2023-06-28
feat: implementing structure and styles of `view` and `projects` components from `projex` in `platform`

AlexPatrie:

* Implements the various `projects` components in `platform`, which is comprised of the project browsing table.
* Implements the `view` component in `platform` which is itself a component of `projects`. This serves as the current "simulation profile page" for `BioSimDB`.
* Adds logic to `ngOnInit` constructor for `app.component.ts` such that a health check can effectively be turned on and off (set to on (`true`) by default).
* Implements Angular `BreakpointObserver` service in various components to listen for screen, and thus, device size.
* Loads `featured.component` images locally, rather than originating from an embedded url.

### DIFF
--- a/apps/platform/src/app/app.component.html
+++ b/apps/platform/src/app/app.component.html
@@ -16,7 +16,7 @@
       </biosimulations-topbar-menu-item>
     </biosimulations-hover-open-menu>
 
-    <biosimulations-hover-open-menu>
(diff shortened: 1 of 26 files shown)


jcschaff (approved):
considering the timelines, I will overlook the any's, but we need to remove these if possible.  Approved.

PR #4725, opened 2023-07-06
fix: s3 import in combine-api and updated create-project unit tests

jcschaff:
Failure when creating a new omex project from a model (e.g. uploaded SBML file).

This functionality in combine-api service defaults to temporarily saving the generated archive in an S3 bucket.  

A bad import statement caused a runtime error when invoking the S3 operation.  This problem was not flagged by MyPy and only manifested itself runtime.  The test coverage for S3 functionality relies on mocks so it wasn't caught there either.

added two explicit test cases, one with the unused 'download' and one that returns a URL to an S3 bucket location.


AlexPatrie (approved):
Looks good!

PR #4733, opened 2023-07-26
feat: changes to `simulators` `dispatch` and `platform` apps

AlexPatrie:

- `chore`: changes pointers in `config.json` for respective repo apps to `.dev`
- `style(ui)`: standardizes the styles from `platform` and `dispatch` in `simulators`.
- `fix`: component behaviors across apps.
- `feat`: added new wording to `home-teaser` component in `platform`


apps/dispatch/src/assets/config.json
```diff
@@ -4,15 +4,15 @@
   "appNameParts": ["run", "Bio", "Simulations"],
   "logo": "assets/images/runbiosimulations-logo/logo.svg",
   "banner": "assets/images/runbiosimulations-logo/logo-white.svg",
-  "appUrl": "https://run.biosimulations.org/",
-  "apiUrl": "https://api.biosimulations.org/",
+  "appUrl": "https://run.biosimulations.dev/",
+  "apiUrl": "https://api.biosimulations.dev/",
   "email": "info@biosimulations.org",
```
jcschaff: still have some hard-coded .dev links

for appUrl, apiUrl, platformAppUrl, platformApiUrl.

libs/config/common/src/lib/endpointLoader.ts
```diff
@@ -83,8 +83,6 @@ export class EndpointLoader {
 
         this.endpointsTemplate.platformApp = dynamicEndpoints?.platformApp || 'https://biosimulations.dev';
 
-        this.endpointsTemplate.dispatchApp = 'https://run.biosimulations.org';
-
```
jcschaff, line 87: why remove only the endpointsTemplate.dispatchApp URLs and not the others?  shouldn't this one be a 'dynamiccEndpoint' also?

libs/config/common/src/lib/endpointLoader.ts
```diff
@@ -18,7 +18,7 @@ export type DynamicEndpoints = { [key in Endpoint]: string | undefined };
 export type LoadedEndpoints = { [key in Endpoint]: string };
 
 export class EndpointLoader {
-  public useDevRuns = false;
+  public useDevRuns = true;
```
jcschaff, line 21: don't forget, this logic should be grabbed from the environment.env configuration variable if possible. ... maybe next PR.

jcschaff (approved):
approved, but with comments.  If you choose you can address them in another PR

libs/config/common/src/lib/endpointLoader.ts
AlexPatrie: Great question. As you can see, I have set the endpoint for dispatch explicitly above (which sets to `.dev`). I then choose to set it with a different value explicitly here. Since we were iterating over the same schema for `endpointsTemplate`, setting a repeated value like this with a method saves code duplication, etc. Only this field is originally targeted as it was pertinent to our concerns, but it is a chore of mine to set all of these fields dynamically.

libs/config/common/src/lib/endpointLoader.ts
AlexPatrie: Correct.

apps/dispatch/src/assets/config.json
AlexPatrie: Thank you for the reminder to change this back to the original value!

PR #4739, opened 2023-08-04
fix: adjustments to repo pointing styles and behavior

AlexPatrie:
_**What does this PR do?**_:
- Addressed the need for route pointer adjustment ahead of the site conversion.
- Addresses styles/behaviors across repo app components as needed.
- Ensure proper dev/prod pointing.


apps/simulators/src/app/app.component.html
```diff
@@ -8,49 +8,44 @@
...
       </biosimulations-topbar-menu-item>
     </biosimulations-hover-open-menu>
 
     <biosimulations-hover-open-menu>
       <biosimulations-topbar-menu-item heading="Utilities" icon="construction" id="trigger">
       </biosimulations-topbar-menu-item>
-
       <ng-container id="items">
-        <biosimulations-dropdown-menu-item heading="Convert a file" [routerLink]="['/utils', 'convert-file']">
+        <biosimulations-dropdown-menu-item heading="Convert a file" [href]="config.utilsConvertFileUrl">
```
jcschaff: I don't understand why you would replace a relative link (which always stays within the app) with a configured absolute link.

libs/shared/assets/src/assets/images/default-resource-images/betaminus.svg
jcschaff, line 1: Is this image used, it is new, but didn't see where it was referenced.  (also marine_illustration.png)

apps/simulators/src/app/app.component.html
AlexPatrie: Ugh. I did not in fact do this intentionally: somehow, just this was copied over in a change from fbd39dc8a37b42d445a9bb03ca0ff9005c27e541, where you adjusted the routing for items within `platform`. This is not applicable (of course) to this app. Fixing now!

libs/shared/assets/src/assets/images/default-resource-images/betaminus.svg
AlexPatrie: This image was originally added to be the value for the `platform.home.featured.service.runSimulation` service on `platform.home -> biosimulations-home-teaser`. I since replaced this image with another but kept the asset. I will tidy up the space now.

jcschaff (approved):
looks good, approved.

PR #4740, opened 2023-08-07
fix: repo qa

AlexPatrie:
_**What does this PR do?**_:
- Adjusts styles across repo components.
- QA as needed.


apps/dispatch/src/app/components/home/home.component.html
```diff
@@ -1,5 +1,5 @@
-<div class="home-container-run">
-  <mat-card class="description-run">
+<div *ngIf="!isMobileHomeDispatch; else mobileViewDispatchHome" class="home-container run">
```
jcschaff: your scss file refers to a 'home-container-run' but here you use 'home-container' and 'run' selectors.

jcschaff (commented):
see comment about scss selectors

apps/dispatch/src/app/components/home/home.component.html
AlexPatrie: I appreciate having an extra set of eyes for something like this!

jcschaff (approved):
looks good, comment about css selector has been resolved.

PR #4742, opened 2023-08-10
feat: cross repo app routing implementation

AlexPatrie:
_**What does this PR do?**_:
To address the concern of cross-repo, inter-app routing:
- Adjusts the attributes of the `ConfigService` object to include only urls which are interpreted as truly 'external'.
- Implements the use of new `AppRoutes` instances in a given app's `app.component.ts` constructor via a global variable. The global variable is exported such that it may be accessed by both the relative app's `AppComponent` declaration and any other submodule component that may require access.


apps/dispatch/src/assets/config.json
```diff
@@ -11,17 +11,10 @@
   "privacyPolicyVersion": 1,
```
jcschaff, line 11: what about appUrl and apiUrl are they exempt from this refactoring?  what are they used for?

apps/dispatch/src/app/app.component.html
```diff
@@ -32,48 +32,14 @@
...
-          heading="Validate a simulator"
-          href="https://github.com/biosimulators/Biosimulators/issues/new?assignees=&labels=Validate%2Fsubmit+simulator&template=ValidateOrSubmitASimulator.yml&title=%5BSimulation+capabilities%5D%3A+"
-          target="githubissues">
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="Suggest a simulator" [routerLink]="['/utils', 'suggest-simulator']">
-        </biosimulations-dropdown-menu-item>
-      </ng-container>
-    </biosimulations-hover-open-menu>
+   <biosimulations-hero-banner-util-button></biosimulations-hero-banner-util-button>
```
jcschaff, line 35: is this intentional, removing navigation to "utilities" in dispatch app?

apps/platform/src/assets/config.json
```diff
@@ -7,28 +7,15 @@
...
-  "utilsValidateSimulationUrl": "https://run.biosimulations.org/utils/validate-simulation",
-  "utilsValidateMetadataUrl": "https://run.biosimulations.org/utils/validate-metadata",
-  "utilsValidateProjectUrl": "https://run.biosimulations.org/utils/validate-project",
-  "utilsSuggestSimulatorUrl": "https://run.biosimulations.org/utils/suggest-simulator",
-  "dispatchApiUrl": "https://api.biosimulations.org/",
   "dispatchNewIssueUrl": "https://github.com/biosimulations/biosimulations/issues/new/choose",
   "dispatchNewPullUrl": "https://github.com/biosimulations/biosimulations/compare",
-  "simulatorsAppUrl": "https://biosimulators.org/",
-  "simulatorsAppBrowseSimulatorsUrl": "https://biosimulators.org/simulators",
   "simulatorsApiUrl": "https://api.biosimulators.org/",
```
jcschaff: shouldn't you remove simulatorsApiUrl here? You removed it in the dispatch app config file.

apps/platform/src/app/app.component.html
```diff
@@ -1,108 +1,75 @@
...
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="Suggest a simulator" [routerLink]="['/utils', 'suggest-simulator']">
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="BioSimulations REST API" [href]="config.apiUrl">
-        </biosimulations-dropdown-menu-item>
-      </ng-container>
-    </biosimulations-hover-open-menu>
-
```
jcschaff, line 64: you removed the utilities navigation from the platform app?  is this intentional?

apps/projex/src/app/app.component.html
```diff
@@ -1,66 +1,32 @@
...
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="Suggest a simulator" [href]="config.utilsSuggestSimulatorUrl">
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="BioSimulations REST API" [href]="config.apiUrl">
-        </biosimulations-dropdown-menu-item>
-      </ng-container>
-    </biosimulations-hover-open-menu>
-
```
jcschaff, line 54: I know that /app/projex is going away soon, but here again the navigation for utilities is removed.  In this case, you could even hard code the *.org (or keep it in the config.json).

apps/simulators/src/app/app.component.html
```diff
@@ -1,89 +1,55 @@
...
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="Suggest a simulator" [routerLink]="['/utils', 'suggest-simulator']">
-        </biosimulations-dropdown-menu-item>
-
-        <biosimulations-dropdown-menu-item heading="BioSimulators REST API" [href]="config.apiUrl">
-        </biosimulations-dropdown-menu-item>
-      </ng-container>
-    </biosimulations-hover-open-menu>
-
```
jcschaff, line 59: removed Utilities navigation from simulators app, is this intentional?  Did move the navigation to the Utilities somewhere else, it seems to be gone from all of the apps ???

jcschaff: it shouldn't be too hard to add these links to the appRoutes class

libs/config/common/src/lib/app-routes.ts
```diff
@@ -2,9 +2,9 @@ import { environment, envs } from '@biosimulations/shared/environments';
 import { EndpointLoader, LoadedEndpoints } from './endpointLoader';
 
 export class AppRoutes {
-  private simulatorsApp: string;
-  private dispatchApp: string;
-  private platformApp: string;
+  public simulatorsApp: string;
+  public dispatchApp: string;
+  public platformApp: string;
```
jcschaff, line 8: I think that you could add methods for the utilities links (when not locally routed) - then you can restore the Utilities navigation safely.

libs/shared/ui/src/lib/hero-banner-util-button/hero-banner-util-button.component.html
```diff
@@ -0,0 +1,21 @@
+<biosimulations-hover-open-menu>
+   <biosimulations-topbar-menu-item heading="Utilities" icon="construction" id="trigger">
+   </biosimulations-topbar-menu-item>
+   <ng-container id="items">
```
jcschaff, line 4: so this new component allows local navigation to the utilities using local routing, but most apps don't have these routes.  So the new design is to (1) navigate to the right app, and (2) then use the utilities of that app?  This could make sense, but maybe it deserves some discussion.

jcschaff (changes requested):
Nice Job, but please address my comments, especially regarding the missing Utilities navigation.

libs/shared/ui/src/lib/hero-banner-util-button/hero-banner-util-button.component.html
AlexPatrie: This new component composition is intended to reduce and prevent code duplication. Notice that each app has a reference to a `utils` submodule. Each relative ui `app.component` already had a reference to `utils` as a navigation button linked with `routerLink`. Given the fact that these `biosimulations-dropdown-menu-item`s are simply re-iterated in each of the ui apps, this is logical solution to prevent the duplication. Using `routerLink` to reference each of the `utils` items in this implementation works as identically as the original implementation, but without the code duplication.

apps/dispatch/src/app/app.component.html
AlexPatrie: This is intentional as there is the new shared component in its place.

apps/platform/src/assets/config.json
AlexPatrie: Yes thank you for the heads up!

AlexPatrie: Thanks for the heads up! I have removed these fields and implemented them as public properties of the new shared component as such:

`public validateSimulatorUrl = 'https://docs.biosimulations.org';`
`public apiUrl = 'https://api.biosimulations.org';`
`public simulatorsApiUrl? = 'https://api.biosimulators.org';`
`public docsUrl = 'https://docs.biosimulations.org';`

apps/dispatch/src/assets/config.json
AlexPatrie: These fields are used as input parameters for `biosimulations-navigation` and have not been a part of this refactor. The logic is that I am housing all links that are truly external in `config.json`.

libs/shared/ui/src/lib/hero-banner-util-button/hero-banner-util-button.component.html
jcschaff: I'm surprised that each app has local implementations of the same utils, is that really the case?

jcschaff (approved):
approved after discussion and some changes.

PR #4744, opened 2023-08-13
fix: removed comments and unused code to appease CI/CD

AlexPatrie:
_What does this PR do?_:
- Refactors nearly all relative `...component.scss` files in both app and shared components.
- Adjusts styling strategy of some components such that code duplication is reduced.
- Removes unused code within comments, import declarations, and single-line comments.
- Transference of scope within certain component/class style declarations.
- Attempts to address a very frequent issue of unterminated string tokens (see repo issues).


jcschaff (approved):
looks good, approved.

PR #4747, opened 2023-08-15
fix: cross-repo content adjustments

AlexPatrie:
_What does this PR do?_:
- `fix`: removes `important!` directive from global font declaration that was overriding implementation styles. Fixes an issue with browsing tables.
- `fix`: corrects svg colors of simulators app logo.
- `fix`: Adds `/projects` to pointing of hero banner nav buttons marked "BioSimDB".
- `chore`: replaces flex layout of browsing table components to implement responsive design.

### DIFF
--- a/apps/dispatch/src/app/app.component.html
+++ b/apps/dispatch/src/app/app.component.html
@@ -35,16 +35,14 @@
     <biosimulations-hero-banner-util-button></biosimulations-hero-banner-util-button>
 
     <biosimulations-hover-open-menu
-      [href]="appRoutes.simulatorsApp"
       [disabled]="true"
-      matTooltip="Browse simulator tools">
-      <biosimulations-topbar-menu-item heading="BioSimulators" icon="link" id="trigger">
-      </biosimulations-topbar-menu-item>
+      [href]="appRoutes.simulatorsApp"
+      matTooltip="Browse simulation projects">
+      <biosimulations-topbar-menu-item heading="BioSimulators" icon="link" id="trigger"> </biosimulations-topbar-menu-item>
     </biosimulations-hover-open-menu>
   </biosimulations-topbar-menu>
 
   <ng-container id="news"></ng-container>
-
   <router-outlet></router-outlet>
 </biosimulations-navigation>
 
--- a/apps/dispatch/src/app/app.component.html
+++ b/apps/dispatch/src/app/app.component.html
@@ -9,7 +9,7 @@
   [healthy]="healthy$ | async">
   <!-- Start Hero Banner -->
   <biosimulations-topbar-menu>
-    <biosimulations-hover-open-menu [disabled]="true" [href]="appRoutes.platformApp">
+    <biosimulations-hover-open-menu [disabled]="true" [href]="appRoutes.platformApp" matTooltip="Browse simulation projects">
       <biosimulations-topbar-menu-item heading="BioSimDB" icon="browse" id="trigger"> </biosimulations-topbar-menu-item>
     </biosimulations-hover-open-menu>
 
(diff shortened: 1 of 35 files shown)


apps/dispatch/src/app/app.component.html
```diff
@@ -34,21 +34,15 @@
...
-    </biosimulations-hover-open-menu>
-
     <biosimulations-hover-open-menu
-      [href]="appRoutes.simulatorsApp"
       [disabled]="true"
-      matTooltip="Browse simulator tools">
-      <biosimulations-topbar-menu-item heading="BioSimulators" icon="link" id="trigger">
-      </biosimulations-topbar-menu-item>
+      [href]="appRoutes.simulatorsApp"
+      matTooltip="Browse simulation projects">
```
jcschaff: think it should be **"Browse simulator tools"** if the link is `appRoutes.simulatorsApp`

apps/dispatch/src/app/components/home/home.component.ts
```diff
@@ -11,6 +11,7 @@ import { DISPATCH_APP_ROUTES } from '../../app.component';
 export class HomeComponent implements OnInit {
   public isMobileHomeDispatch = false;
   public appRoutes = DISPATCH_APP_ROUTES;
+  public logoPath = 'assets/images/runbiosimulations-logo/logo.svg';
```
jcschaff, line 14: general comment - hard coded asset paths - does the home.component.spec.ts fail if it can't find the file?   This is more of a future looking comment than a request for change.

apps/platform/src/assets/config.json
```diff
@@ -7,7 +7,6 @@
   "appUrl": "https://biosimulations.org/",
   "apiUrl": "https://api.biosimulations.org/",
   "email": "info@biosimulations.org",
-  "docsUrl": "https://docs.biosimulations.org/",
```
jcschaff, line 10: maybe we still need this (for use in hero-banner-util-button.component.html)??? ... or else, where does the hero button get the "docsUrl" config?

libs/config/angular/src/lib/config/config.service.ts
```diff
@@ -49,7 +10,6 @@ export class ConfigService {
   appUrl!: string;
   apiUrl!: string;
   email!: string;
-  docsUrl?: string;
```
jcschaff, line 52: if the common hero-banner-util-button.component.html still refers to [href]="docsUrl", don't we still need it in each app's config?

jcschaff (changes requested):
please take a look at my comments, I think that a link is wrong, and I'm wondering if the docUrl config is correct.

libs/config/angular/src/lib/config/config.service.ts
AlexPatrie: I have moved the setting logic out of the scope of `Config` and into a property of the shared `biosimulations-hero-util-button` component itself as such (.ts):

`public validateSimulatorUrl = 'https://docs.biosimulations.org';`
`public apiUrl = 'https://api.biosimulations.org';`
`public simulatorsApiUrl? = 'https://api.biosimulators.org';`
`public docsUrl = 'https://docs.biosimulations.org';`

and implemented in the component's html as such (for example):

<biosimulations-dropdown-menu-item heading="Validate a simulator" [href]="validateSimulatorUrl">
</biosimulations-dropdown-menu-item>


Since the `docsUrl` value is used ONLY for setting a value in the nav button on the hero banner and is the same for each app, this was done to prevent duplication.

apps/dispatch/src/app/app.component.html
AlexPatrie: Yes!! Thanks!

apps/dispatch/src/app/components/home/home.component.ts
AlexPatrie: Great question. Currently the relative `dispatch.home.component.spec.ts` uses `compileComponents()` to construct the component from the ancestral templates/styles and validates construction with:

` it('should create', () => {`
`    expect(component).toBeTruthy();`
 ` });`

So to answer your question, the test does not explicitly look for this file but will indeed fail if unable to get/set the value.

jcschaff: cool, addressed !!!

libs/config/angular/src/lib/config/config.service.ts
jcschaff: good answer, resolved!

apps/platform/src/assets/config.json
jcschaff: resolved - moved this configuration to the appropriate 'hero' button config.

jcschaff (approved):
comments were addressed, approved!

PR #4748, opened 2023-08-18
fix: adjusted hero banner color for simulators app to align with logo color

AlexPatrie:
_What does this PR do?_:
- To fix the issue of incorrect rendering (too-dark) of the `biosimulators` logo, color changing logic is introduced in the `biosimulations-navigation` component to be of `background: var(--accent-color);` if `appName === 'BioSimulators'`.


libs/shared/ui/src/lib/hero-banner-util-button/hero-banner-util-button.component.ts
```diff
@@ -1,11 +1,13 @@
...
+import { Component, Input } from '@angular/core';
 
 @Component({
   selector: 'biosimulations-hero-banner-util-button',
   templateUrl: './hero-banner-util-button.component.html',
   styleUrls: ['./hero-banner-util-button.component.scss'],
 })
 export class HeroBannerUtilButtonComponent {
+  @Input()
+  public appName!: string;
```
jcschaff, line 10: This is probably out of scope and unimportant here - just a comment about style and type safety.  I'm not asking you to make the following change - but consider it for the future.

I'm not crazy about appName being a string with string literals in the corresponding HTML (if we change app names or misspell).  I was thinking about typing this as a union of literals (e.g. "BioSimulators" | "BioSimulations" | ...).

Potential change

in config.ts (or wherever the config typing is defined):
`type AppName = "BioSimulators" | "BioSimulations" | ..`. 
`appName: AppName;`

then here appName string compares will be checked by the compiler (I think)
`appName: AppName;`

jcschaff (approved):
I left one comment which about safety of string literals.  A true commentary, not asking you to make the corresponding change - but think about it for the future.
Approved!

libs/shared/ui/src/lib/hero-banner-util-button/hero-banner-util-button.component.ts
AlexPatrie: I like your approach of having a mapping rather than a metaphysically ternary operation: we currently have no way of validating the string literal in the above change and original implementation.

PR #4755, opened 2023-08-26
fix: disable COBRApy and RBApy deploy tests until new Gurobi license

jcschaff:
**What new features does this PR implement?**
The Gurobi license has expired at the HPC facility which is hosting simulations for dev and production sites.  Two solvers, `COBRApy` and `RBApy` need `Gurobi` to execute properly.  

This PR removes the COBRApy and RBApy simulations from the 'dev' site acceptance tests, which is currently blocking production deployments.  These tests are executed via .github/workflows/deploy.yml in the job entitled "Test the development deployment" during the continuous deployment Github Action.

Note: Once the new `Gurobi` 10.0 license is obtained, and the `biosimulators_cobrapy` and `biosimulators_rbapy` container images are rebuilt and submitted as new versions 

**added manual checkpoint - uncommented github environment "review_dev_deployment_tests_environment" to enforce production deployment approval by Bilal, Jonathan, Ion or Jim.

**What bugs does this PR fix?**
the "Test the development deployment" job in the continuous deployment GitHub action is broken due to a Gurobi license error - this PR will temporarily bypass that failure.

PR #4756, opened 2023-08-28
fix: force build test deploy workflow for merged PR 4755

jcschaff:
forces build/deploy workflow for previous PR #4755 which changed only a text file which was only used in continuous deployment actions.

Here, we introduce a small text change in a web page to force build/test/deploy workflow


AlexPatrie (approved):
Looks good.

PR #4758, opened 2023-08-29
fix: update hsload args for h5pyd 0.14.1 to upload to HSDS from Slurm

jcschaff:
update hsload arguments for new version of h5pyd package which was upgraded on the HPC server from 0.8.4 to 0.14.1 (to correspond to currently deployed version of HSDS.

username option changed to user


AlexPatrie (approved):
Looks good. Just as a heads up: GitHub is griping about using `any` (as per an earlier conversation we had) which, in my experience, may trigger the `Unterminated string token` error in pipeline. Re-running the failed job should do the trick to pass :)

PR #4759, opened 2023-08-29
fix: removed other switch

jcschaff:
another switch change "--retries 10" is broken in 0.14.1 (throws exception in HTTP layer in hsload)


AlexPatrie (approved):
Approved!

PR #4760, opened 2023-08-29
fix: HSDS upload HPC Jobs to retry randomly many times upon failure

jcschaff:
Bug: simulation jobs often fail to run when submitted simultaneously because of contention for HSDS REST API.

Logic added in Slurm job to try to upload HSDS data up to 40 times, sleeping randomly from 5 to 15 seconds between each try.


AlexPatrie (approved):
Awesome.

PR #4784, opened 2023-12-29
fix: adjusted overflow for side bar content

AlexPatrie:
_What does this PR do?_:
- Fixes card width in simulation run text page sidebar component


jcschaff (approved):
simple CSS change, looks good to me.

PR #4786, opened 2024-01-03
fix: updated structured log to include more extensive message

AlexPatrie:
_What does this PR do?_:

- Adds `stepsInfo.reason` to `dispatch-service.submission.complete.processor.CompleteProcessor.getFailedStepErrorMessage` error message for more verbose output.

### DIFF
--- a/apps/dispatch-service/src/app/submission/complete.processor.ts
+++ b/apps/dispatch-service/src/app/submission/complete.processor.ts
@@ -172,7 +172,7 @@ export class CompleteProcessor extends WorkerHost {
   }
 
   private getFailedStepErrorMessage(step: stepsInfo, failedSteps: stepsInfo[]): string {
-    let message = step.description + ' ... failed';
+    let message = step.description + ' ... failed due to ' + step.reason;
     let failedDueToChild = false;
     const children = step.children;
     children.forEach((child) => {
--- a/apps/dispatch-service/src/app/submission/complete.processor.ts
+++ b/apps/dispatch-service/src/app/submission/complete.processor.ts
@@ -257,7 +257,7 @@ export class CompleteProcessor extends WorkerHost {
         returnValue: flow.job.returnvalue.data,
         required: flow.job.data.required,
         description: flow.job.data.description,
-        errorMessage: flow.job.data.errorMessage,
+        errorMessage,
         data: flow.job.data,
         children: children,
       });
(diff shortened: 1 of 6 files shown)


apps/dispatch-service/src/app/submission/complete.processor.ts
```diff
@@ -253,7 +257,7 @@ export class CompleteProcessor extends WorkerHost {
         returnValue: flow.job.returnvalue.data,
         required: flow.job.data.required,
         description: flow.job.data.description,
-        errorMessage,
+        errorMessage: flow.job.data.errorMessage,
```
jcschaff: errorMessage local var already was prefixed with `flow.job.data.errorMessage` (see line 245).  This fix seems to just remove the extra info from `flow.job.data.moreInfo` (line 247) or `flow.job.data.validator` (line 250).

do you really want to remove info here?

apps/dispatch-service/src/app/submission/process.processor.ts
```diff
@@ -145,7 +144,8 @@ export class ProcessProcessor extends WorkerHost {
       data: {
         runId: job.data.runId,
         description: 'Post the SED-ML file to the API',
-        moreInfo: undefined,
+        moreInfo: job.data.moreInfo,
+        reason: job.failedReason,
```
jcschaff, line 148: I cannot find where this field is ever read.  Part of untyped message body, but the message processor for JobQueue.sedmlPost seems to ignore this field.

Does adding this reason field have any effect?

jcschaff (changes requested):
please respond to comments

apps/dispatch-service/src/app/submission/complete.processor.ts
AlexPatrie: Not at all: this seems to have gotten removed on accident. Thanks!

AlexPatrie: I see more of what happened here. `errorMessage` was indeed already locally defined.

apps/dispatch-service/src/app/submission/process.processor.ts
AlexPatrie: I am changing this back to undefined for `moreInfo` for the sake of continuity with the rest of the declarations, but I do think we should eventually implement such fields here so that the post SEDML job and really all the jobs have such metadata. As for `reason`, the same.

jcschaff: okay, I understand.  You wanted to put it there towards a better future design.  One thing that bothers me is that the entire data field is defined as type `any`, and we never define types for the payload of each message/job.  In the future, we should define the type of object which goes into the data field (which may be job specific, or not).

approved.

jcschaff (approved):
requested changes addressed, approved!

PR #4792, opened 2024-01-16
feat: integrated simularium file generation in combine-api

AlexPatrie:
_What does this PR do?:_
- integrates `biosimulators-simularium` as a dependency in `apps/combine-api`.
- configures `combine-api/pyproject.toml` accordingly for versioning.
- introduces a "standard" format for representing the output of spatial simulations with VTK (see [the biosimulators-simularium tooling](https://github.com/biosimulators/Biosimulators_simularium) for more info)
- implements `.simularium` and `.vtk` file generation for Smoldyn (spatial) simulation runs via `biosimulators_simularium.execute()` within the body of `combine_api.handlers.run.handler()`. 
- includes aforementioned file generation in handler output for archive bundling and the creation of `output.zip` for all simulation runs that include Smoldyn.
- introduces appropriate testing logic for the simularium/vtk integration.
- performs minor CSS/style maintenance and adjustment for components throughout repo (fixes nav-button flicker, for example.)

_How did you test the changes?:_
- introduced a new unittest method for smoldyn urls only.
- pytest
- CI/CD


apps/combine-api/tests/fixtures/Min1.txt
```diff
@@ -0,0 +1,117 @@
+# File Min1.txt, by Steve Andrews, 11/2009.
```
jcschaff: Is this file referenced anywhere? doesn't seem like it.

apps/combine-api/clear-cache.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/zsh 
+
+if yes | poetry cache clear PyPI --all && yes | poetry cache clear _default_cache --all; then
```
jcschaff: does this condition ever fail?  doesn't 'yes' always win?

apps/combine-api/delete-env.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
jcschaff: can we remove this file and install-deps.sh, install-smoldyn.sh, job.sh - or add the commands to a 
 ... or add a README-dev.md with helpful commands if you think that they are generally useful.  same goes for

jcschaff (approved):
fixed my comments, looks good.

jcschaff:
do you have to merge master in first? it seems to be using the master workflow scripts, and not finding simdata-api??

PR #4794, opened 2024-01-26
fix: simdata-api deployment

jcschaff:
resuses the existing path data.biosimulations.dev/org which was used by HSDS for simdata-api.  This allows the TLS cert to be reused.

a cert error occurs when accessing the api externally.


AlexPatrie (approved):
Looks good. This needed to be done.

PR #4795, opened 2024-01-26
fix: disable broken UpdateSitemaps step of Continuous integration

jcschaff:
disable UpdateSitemaps step of Continuous integration action in cd.yml.  It is failing in a non-recoverable way - it bootstraps off of a healthy biosimulations.org site, but since biosimulations.org is not healthy at the moment, this step blocks any progress.

This step will be restored in the future - hopefully in a way that doesn't reach out to the server - or make it optional

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -24,11 +24,11 @@ jobs:
     with:
       branch: dev
 
-  UpdateSitemaps:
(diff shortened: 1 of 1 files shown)


AlexPatrie (approved):
Looks good. Just a heads up that you might encounter autoformatting or errors from the Spell checking action.

PR #4796, opened 2024-01-26
fix: disable updatesitemaps again

jcschaff:
see #4795 - redoing previous PR to disable Update Site Maps prior to release (chicken or egg problem) - broken production site breaks UpdateSiteMaps action, so can't fix production site.


AlexPatrie (approved):
Nice! Just as per what we discussed.

PR #4798, opened 2024-01-29
fix: look for reports.h5 in both contents and outputs S3 paths

jcschaff:
- some datasets only have the simulation results in `files.biosimulations.dev/simulations/{run_id}/contents`, others have it only in `files.biosimulations.dev/simulations/{run_id}/outputs`, some have reports.h5 in both locations.  simdata-api should be able to support both conventions.
- renamed path for simdata-api from data.biosimulations.dev to simdata.api.biosimulations.dev for symmetry with other secondary services (e.g. combine-api).  Was temporarily named data.biosimulations.dev to reuse networking from hsds service which was recently replaced.
- fixed root logger configuration in simdata-api so that INFO and above will be logged - we may revisit in the future after simdata-api becomes more mature.

simulation runs which stored their reports.h5 in 'outputs' subdirectory were not supported.

altered unit testing in simdata-api to verify simulation data support for both 'contents' and 'outputs' subdirectories.

### DIFF
--- a/apps/simdata-api/simdata_api/datastore.py
+++ b/apps/simdata-api/simdata_api/datastore.py
@@ -16,11 +16,13 @@ from simdata_api.internal.tensorstore_utils import write_ts_metadata_root, write
 
 logger = logging.getLogger(__name__)
 
+STORE_DIR_NAME = "store"
+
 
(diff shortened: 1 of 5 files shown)


mpw6 (approved):
approved per request from Jim

mpw6 (approved):

mpw6 (approved):

mpw6 (approved):

PR #4804, opened 2024-03-21
feat: add simulation edit functionality to dispatch component within platform

### LINKED ISSUE #4753: 🐛:  'run' page doesn't allow steady state substitutions
### Location

runBiosimulations

### What happened?

There is now a mechanism in KiSAO (and libkisao) to allow steady state algorithm substitutions that did not use to exist.  Any simulator using the latest libkisao can now run a SED-ML file with one steadystate algorithm requested, and use a different one instead.

However, the front end (https://run.biosimulations.org/runs/new) doesn't know that, and requires a strict match to populate the 'simulation tool' list.  It also needs to be hooked up to the new libkisao rules (I assume?) to allow substitutions to be made.

The API runs things just fine, and performs no checks on what simulation tool can be used with what SED-ML algorithm.  See, for example:

https://run.biosimulations.org/runs/64e531906501296249684a11

which is a tellurium run of biomodel 498.  If I hit 'rerun':

https://run.biosimulations.org/runs/new?projectUrl=https:%2F%2Fapi.biosimulations.org%2Fruns%2F64e531906501296249684a11%2Fdownload&simulator=tellurium&simulatorVersion=2.2.8&runName=BIOMD0000000498%20(rerun)

it doesn't allow you to select 'tellurium' again.

### Version

Production (Deployed on a .org domain)

### LINKED ISSUE #4767: 'simulation specification' tab probably lying about having no image to display
https://run.biosimulations.org/runs/64e5319b548ceb84f3ad843c

On the 'simulation specification' tab, it claims there's no image to display:

image

However, I'm almost certain there are indeed images to display there.

### LINKED ISSUE #4769: 'run a new simulation' page needs to be redesigned
https://run.biosimulations.org/runs/new

The old page and the new page have the same problems; this is a proposed improvement, not a new thing introduced by the redesign.

Basically, there's entirely too much space devoted to options that nobody will usually care about.  There are exactly two things that the user should be asked about, and they should be at the very top:

* The OMEX file itself
* What simulator they want to use for it.

That's it!  

The 'Capabilities required' section is auto-filled when you submit the OMEX file, and as far as I know can't be changed anyway, nor should it be.

The 'Simulation tool' section should pull up the appropriate simulators to the top of the list, or just drop the other ones entirely.  The code here sometimes gets it wrong as to what simulators can simulate what (see #4753), but as-is, it doesn't let you select the greyed-out simulators anyway.  Don't list the greyed-out simulators at all; just let the user choose among the allowed ones.

The 'Commercial solvers' section should only appear if a simulator was chosen that might need it.  Even then, it should be a small checkbox with a pop-up explanation; not something that takes 1/4 of the page.

'Computational Resources' should normally be hidden, and only expanded if requested.  I've never once needed to change these.

'Metadata':  While you could allow the user to fill in a name here, it shouldn't be required.  By default, just assign an arbitrary unique name, perhaps 'timestamp/OMEX filename'.  

Notifications:  Another section that can normally be skipped.  Also, if a user fills out their email address, that should be consent enough; they shouldn't need to additionally check the 'I am OK receiving an email' box.  Just tell them (maybe in a popup) that the address will only be used to email them a single notification when the simulation is complete.  Also, I would tie this to the user's account, and let them set a default send/don't_send option so they don't have to select it every time if they want it.

The 'Run' button shouldn't be only at the bottom; it should be accessible from the top where you can select your two main options.

AlexPatrie:
_What does this PR do?:_
- (feat): implements service methods for `combine-api` within `platform/dispatch` to render all available editable simulation parameters.

- (feat): creates and implements a new `shared-simulation-service` within common `./libs/shared/services` to handle custom simulation re-run.

- (fix): refactors structure of dispatch form to hide less-used configurations in an "Advanced Settings" `mat-expansion-panel` ( closes: #4769 )

- (fix): opens the `minPolicy` attributes to any level of algorithm substitution policy for the instantiation of `KisaoAlgorithmSubstitution` objects within `platform/dispatch.component.ts` ( closes: #4753 )

- (fix): fixes project metadata get handler for all images associated with a given simulation run, and thus COMBINE/OMEX archive within `biosimulations-project-metadata.component` ( closes: #4767 )

_How was this PR tested?:_
- pytest

- `spec.ts` for TS content


jcschaff (approved):
looks good, approved

PR #4841, opened 2024-10-02
chore: redirect for run biosimulations to point to biosimulations and GUI updates

AlexPatrie:
_What does this PR do?:_
- Adds a redirect for `https://run.biosimulations.dev` to point instead to `https://biosimulations.dev`
- Adds the same aforementioned redirect for "production" (`.org`).
- Adds default values to form directive configurations in `create-project.component` within `libs/shared`.
- Adds default values for x and y axis in 2d line plot creation within `libs/simulation-runs/ui/src/lib/design-line-2d-viz/design-line-2d-viz.component` as well as minor UI (style) refactors, removing unused data headers.
- Minor UI (style) refactors as needed.

### DIFF
--- a/apps/platform/src/app/app.component.ts
+++ b/apps/platform/src/app/app.component.ts
@@ -6,6 +6,9 @@ import { UpdateService } from '@biosimulations/shared/pwa';
 import { AppRoutes } from '@biosimulations/config/common';
 import { BreakpointObserver, Breakpoints } from '@angular/cdk/layout';
 import { Observable } from 'rxjs';
+import { MatDialog } from '@angular/material/dialog';
+import { CookieConsentComponent } from '../../../../libs/analytics/angular-analytics/src/lib/cookie-consent/cookie-consent.component';
+import { cookieConsentType } from '../../../../libs/analytics/angular-analytics/src/lib/datamodel';
 
 export const PLATFORM_APP_ROUTES = new AppRoutes();
 @Component({
@@ -20,10 +23,12 @@ export class AppComponent implements OnInit, AfterViewInit {
   private checkForHealth = true;
   public mobileLinkTarget = '_blank';
   public mobileLink?: string;
+  public debugCookies = true;
   private _appRoutes: AppRoutes;
 
   public constructor(
     public config: ConfigService,
+    public dialog: MatDialog,
     private scrollService: ScrollService,
     private updateService: UpdateService,
     private healthService: HealthService,
@@ -46,10 +51,22 @@ export class AppComponent implements OnInit, AfterViewInit {
       console.log(this.healthy$);
     }
     this.checkClientScreenSimulations();
+
+    if (this.debugCookies) {
+      this.openCookieConsentDialog();
+    }
   }
 
   public ngAfterViewInit(): void {
     this.scrollService.init();
+  }
+
+  public openCookieConsentDialog(): void {
+    this.dialog.open(CookieConsentComponent, {
+      width: '625px',
+      position: { bottom: '1.5rem', left: '0' },
+      panelClass: 'cookies-container',
+    });
   }
 
   private checkClientScreenSimulations(): void {
--- a/apps/platform/src/app/app.component.ts
+++ b/apps/platform/src/app/app.component.ts
@@ -7,8 +7,6 @@ import { AppRoutes } from '@biosimulations/config/common';
 import { BreakpointObserver, Breakpoints } from '@angular/cdk/layout';
 import { Observable } from 'rxjs';
 import { MatDialog } from '@angular/material/dialog';
-import { CookieConsentComponent } from '../../../../libs/analytics/angular-analytics/src/lib/cookie-consent/cookie-consent.component';
-import { cookieConsentType } from '../../../../libs/analytics/angular-analytics/src/lib/datamodel';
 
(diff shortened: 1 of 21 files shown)


apps/platform/src/app/app.component.ts
```diff
@@ -24,6 +25,7 @@ export class AppComponent implements OnInit, AfterViewInit {
 
   public constructor(
     public config: ConfigService,
+    public dialog: MatDialog,
```
jcschaff: doesn't seem to be used, am I missing something?

libs/analytics/angular-analytics/src/lib/cookie-consent/cookie-consent.component.ts
```diff
@@ -20,10 +20,10 @@ import {
 })
 export class CookieConsentComponent {
   public cookies: Cookie[] = [
-    new RequiredCookie(),
+    // new RequiredCookie(),
```
jcschaff: are you sure we should comment out tracking and required.  our advisors suggested that we need to track usage ... does this remove this ability in the future ... maybe we should keep it for now.

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/uniform-time-course-simulation.component.ts
```diff
@@ -44,10 +45,11 @@ export class UniformTimeCourseSimulationComponent implements IFormStepComponent,
   }
 
   public populateFormFromFormStepData(formStepData: FormStepData): void {
-    this.formGroup.controls.initialTime.setValue(formStepData.initialTime);
-    this.formGroup.controls.outputStartTime.setValue(formStepData.outputStartTime);
-    this.formGroup.controls.outputEndTime.setValue(formStepData.outputEndTime);
-    this.formGroup.controls.numberOfSteps.setValue(formStepData.numberOfSteps);
+    // This method is called onInit and thus should set the default vals
```
jcschaff, line 48: why is this needed if the default values are set in the contructor?  will this play nicely if the user wants to override the defaults?

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/upload-model-component.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('UploadModelComponent', () => {
     fixture.detectChanges();
 
     testFormats = [
-      createTestOntologyTerm('test format 1'),
+      createTestOntologyTerm('test format_2585'),
```
jcschaff: is this intentional or a debugging edit?

jcschaff (changes requested):
please take a look at the comments

apps/platform/src/app/app.component.ts
AlexPatrie: This was a debugging artifact which I have now removed.

libs/analytics/angular-analytics/src/lib/cookie-consent/cookie-consent.component.ts
AlexPatrie: Another debugging artifact that was not originally reverted. Now it is uncommented!

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/uniform-time-course-simulation.component.ts
AlexPatrie: Great question. In angular forms, there is a difference between setting a default value (that which is specified in `constructor`) and setting a form value such that it actually renders. The default value specified in this component's constructor is only collected `onSubmit()` (called by the parent component), and thus will not render on the screen as a `Change` in the form data. This call to `.setValue()` originally set the data from the `formStepData` of the previous step, which was `null`. The form data source is actively subscribed to user changes, so this explicit setting will not mess with new override specifications.

TLDR; Explicitly setting these values is required for rendering the default values and will not tamper with user specifications.

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/upload-model-component.spec.ts
AlexPatrie: This is another matter relating to default values in angular. `format_2585` is the tag for the non-spatial continuous framework option which we discussed should be the default value for the modeling framework option in `create-project/upload-model-component`. The format of this test requires this value.

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/uniform-time-course-simulation.component.ts
jcschaff: thanks, great explanation.

libs/simulation-project-utils/simulation-project-utils/src/lib/ui/create-project/form-steps/upload-model-component.spec.ts
jcschaff: okay thanks.

jcschaff (approved):
Thanks for answering my questions, approved!

PR #4857, opened 2024-12-02
fix: correct log from simdata api client wrapper module

jcschaff:
corrects erroneous log message when using Simdata module in NestJS.   Very safe.


AlexPatrie (approved):

luciansmith (approved):

luciansmith:
FWIW, the review panel complains about 'only.txt' not having a newline on its own at the end.

PR #4859, opened 2024-12-03
feat: implement data caching in dispatch

AlexPatrie:


jcschaff (approved):
looks good, please remove anything at all risky in the short term.  Approved.

PR #4861, opened 2024-12-12
fix: use lowercase containers in simulator container URLs

jcschaff:
The Virtual Cell simulator container image URL was generated from the simulator `id` as defined in its biosimulator.json which was `VCell` rather than the lower case `vcell`.  For the recent submission of VCell 7.7.0.12, this created a docker URL used by singularity of docker://ghcr.io/biosimulators/VCell:7.7.0.12, which is rejected with the following message:

> While making image from oci registry: error fetching image to cache: failed to get checksum for 
> docker://ghcr.io/biosimulators/VCell:7.7.0.12: unable to parse image name docker://ghcr.io/biosimulators/VCell:7.7.0.12: 
> invalid reference format: repository name must be lowercase

now, when constructing the URL of a simulator docker image, the simulator name is always forced to be lowercase.  VCell may or may not be the only simulator whose `id` is mixed case, this fix should handle all cases.  There are other places in the code which compare the "normalized" simulator ids, accounting for mixed case.

This fixes the problem where the Virtual Cell simulator jobs timeout after being recently deployed, because the singularity image pre-generation fails with the message above.

also, improved a logging message when the api connects to Redis.

this has not been tested, but is very unlikely to do any harm.

### DIFF
--- a/apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts
+++ b/apps/dispatch-service/src/app/services/sbatch/sbatch.service.ts
@@ -58,7 +58,7 @@ export class SbatchService {
     const storageKey = this.configService.get('storage.accessKey');
     const storageSecret = this.configService.get('storage.secret');
 
-    const simulatorImage = `docker://ghcr.io/biosimulators/${simulator}:${simulatorVersion}`;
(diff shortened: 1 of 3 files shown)


AlexPatrie (approved):
Looks good. Approved.

PR #4862, opened 2024-12-20
feat: remove /run endpoints from combine-api

jcschaff:
this PR removes the /run/run and /run/simulators endpoints from `combine-api`.  These endpoints were not used in production and contribute to the overall complexity of combine-api service as well as the `Angular` and `NestJS` language bindings.

since this is merely a removal of function, if the typescript builds for `Angular` and `NestJS` then there was not dependency. the endpoints and corresponding JSON Schema definitions were removed from the OpenAPI `spec.yml.`  The corresponding handlers were removed from `combine-api` service.


AlexPatrie (approved):
Looks good! This is an important step towards maintainability.

